PR #41, opened 2014-11-10
Environment-injected Base URL

### LINKED ISSUE #32: Create a facility to support multiple environments
During the primary build of a project there are often multiple domains basepaths. Development, QA, Demo, Production, Feature-branch instances, and so on. Any of these might need meaningful behat and other sorts of testing processes.

We need a method to make the basepath adapative or swappable via configuration. Making basepath a default configuration with overrides from command-line might be the most natural, for example: ```grunt --env=grayside.example.com``` or ```GRUNT_BASEPATH=grayside.example.com```. The latter leverages environment parameters, which may allow for automating how this happens without explicit developer action.

grayside:
This PR adds a new `domain` key to facilitate shipping a default environment parameter. When used alongside siteUrls with the appropriate configuration templating, this allows absolute URLs to each site to be adaptive. Domain can be specified in the `Gruntfile.json` for your project, but an environment variable `GRUNT_DRUPAL_DOMAIN` will take precedence.

A new utility module is created `lib/gdt.js` which can be loaded with `var GDT = require('lib/gdt');`. This allows test-oriented tasks to submit a list of paths and get a set of absolute URLs for a particular baseUrl. `GDT.expandUrls(grunt.config('config.domain'), ['/', 'admin'])`

This change does not modify documentation, I thought to leave that for the first follow-up leveraging the system.

This PR fixes #32.


arithmetric:
This looks good for the 0.7.0 release. To align with the env var name changes in #141, I'll add a change to the env var prefix.

Issue #42, opened 2014-11-10
Add support for CSS/SASS and JS lint


arithmetric:
Help wanted to identify the best tools and sensible defaults for the lint settings.

CSS:
https://github.com/gruntjs/grunt-contrib-csslint
http://csslint.net/

SASS:
https://github.com/ahmednuaman/grunt-scss-lint
https://github.com/causes/scss-lint

JS:
https://github.com/gruntjs/grunt-contrib-jshint
https://github.com/stephenmathieson/grunt-jslint
https://github.com/reid/node-jslint
https://github.com/jshint/jshint

grayside:
We may want to coordinate the rules with @evanlovely given https://github.com/phase2/pattern-lab-starter/blob/master/.jshintrc. Alternately, we could see about excluding themes from this kind of quality check along the lines of the theme proxy branch in progress, but that would complicate style violation statistics for the codebase.

EvanLovely:
@grayside @arithmetric I think I realized how best to tell a theme/sub-grunt to run commands: `npm run` and the `scripts` object of `package.json`. So for this current discussion, GDT would `cd` to the theme and run `npm run qa` and in the `package.json` of that theme it would contain:

```json
{
  "scripts": {
    "qa": "grunt scss-lint && grunt css-lint && grunt jshint"
  }
}
```

A big thing to consider is what to do about code quality standards for custom modules and other things not inside the theme. My big litmus test for whether something should be in GDT or the theme is if the theme can stand alone and run Pattern Lab, Scss, & JS without the Drupal backend or GDT so Front End can get started before Back End if needed.

grayside:
Hey Evan,

We already have the proxying of commands to the theme working in a PR. I
don't like relying on package.json for that, but happy to discuss in the PR.

Sounds like you're thinking proxying-enabled themes shouldn't have their
syntax highlighting handled by GDT. This tends to imply theme-specific
standards. Do you think module js and theme js should be allowed to conform
to different practices?

PR #61, opened 2014-11-23
#42 Add ESLint support with grunt-eslint [WIP]

mdeltito:
In reference to #42 (and https://github.com/phase2/grunt-drupal-tasks/issues/42#issuecomment-64136831), this is what ESLint support might look like. The included .eslintrc is from https://www.drupal.org/node/1955232.

Note that these example settings may not be the "best" for validating JS in the context of what grunt-drupal-tasks is trying to provide. They are agreed upon for D8, but whether or not they should be the suggested defaults for all Drupal projects is probably up for debate.

### DIFF
--- a/example/.eslintrc
+++ b/example/.eslintrc
@@ -1,0 +1,32 @@
+{
+  "env": {
+    "browser": true
+  },
+  "globals": {
+    "Drupal": true,
+    "drupalSettings": true,
+    "domready": true,
+    "jQuery": true,
+    "_": true,
+    "matchMedia": true,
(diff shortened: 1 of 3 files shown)


grayside:
This issue needs
* Updates to the D8 rules
* D7 rules
* Incorporation of eslint task into validate task.
* Test coverage to at least demonstrate the successful operation of eslint as implemented

arithmetric:
Pulled this work into a branch in this repo to resolve the merge conflict and make some updates.

This work will continue in #160.

PR #89, opened 2014-12-26
Only load plugins when they are needed.

grayside:
As it is, plugin loading represents only 3-6ms, however this is still a best practice and hopefully will remind us in the future to load grunt plugins only when we're sure we'll need them.


arithmetric:
I did a rebase on this branch with master to pull in the Compass/Bundler tests, and the tests passed on my local.

I also reviewed the approach, and see it works by defining custom tasks with registerTask() where you can then load dependencies. Your patch uses this approach in places we're already using registerTask, rather than an overhaul of all plugins.

I'd like to keep this open for now to consider code style implications and how we should standardize.

PR #93, opened 2014-12-30
Move default PHPCS standard path into code.

grayside:
This and #92 can be merged in any order.

### DIFF
--- a/example/Gruntconfig.json
+++ b/example/Gruntconfig.json
@@ -19,7 +19,6 @@
   },
   "phpcs": {
     "path": "bin/phpcs",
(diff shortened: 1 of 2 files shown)


grayside:
I looked into detecting if the PHP package is present as a way of keying activating the tools, and that seems fine. However, the problem with making an implicit default of `bin/phpcs` is that it's kind of specific to our example composer.json file, and it's something I think is a messy practice for projects that might have custom scripts. Should we remove the composer.json bin-dir setting as part of this?

arithmetric:
I'm going to merge this, and we'll follow up on the question of binary/script paths in #102.

PR #94, opened 2014-12-30
#78: Bump drush make concurrency to 16

### LINKED ISSUE #78: Bump the concurrency for drush make
I.e., include something like --concurrency=16 as an arg for drush make. Is there any reason we can't do that? 

This is a drastic speed increase obviously but I wasn't sure if it was left out for a reason. For example, in the past I had heard that concurrency can create race conditions in cases where different make files specify different versions of modules (such as when you override the version of something that Panopoly ships with from a custom make file), but I thought that had been worked out now.

mikecrittenden:
Fixes #78.

### DIFF
--- a/example/Gruntconfig.json
+++ b/example/Gruntconfig.json
@@ -26,7 +26,7 @@
   },
   "drush": {
     "make": {
-      "args": []
+      "args": ["--concurrency=16"]
     }
   },
   "behat": {


arithmetric:
Thanks @mikecrittenden. I'll merge this change to the example Gruntconfig.json, which can be changed or removed as needed on a project.

@grayside and I discussed adding a more generic and universal method for identifying optimal concurrency, which I'll open a separate issue to track.

PR #96, opened 2014-12-31
Adding output about disk usage to Travis script

arithmetric:


arithmetric:
By default, Travis hides output from commands outside of the main "script" section. However, you can click on an arrow next to the command line to display the output.

See:
Code checkout disk usage

Example working copy disk usage

Here's the current disk usage stat:
- Code checkout: .5mb
- Example working copy (after `npm install`): 18mb
- Fully built (after `grunt` including `composer install`): 74mb

grayside:
Note that it's much harder to expand and understand on mobile :)

PR #97, opened 2014-12-31
Add Zombie.js and driver for JavaScript tests

arithmetric:


grayside:
What's needed here to feel good about merging? I don't have a rig to test zombie integration.

arithmetric:
@grayside I've tested this locally using the following procedure:

1. `test/create_working_copy.sh`
2. `cd test/working_copy/`
3. `grunt`
4. `grunt drush:liteinstall`
5. `grunt drush:runserver &`
6. `grunt test`

The included example test feature includes a scenario that requires JavaScript, and this scenario is now included by the example Behat config. On my local, all running test scenarios pass.

grayside:
I just set up a new environment, so my results might be suspect, but I'm getting issues with composer.
http://i.gyazo.com/a5bf293d21d2e98ca5b68c70bc2e18a0.png

PR #100, opened 2015-01-03
More flexible and granular customization of the compile-theme task [WIP]

grayside:
Been thinking about how the compile-theme task is likely to need customization for the specific needs of many themes. While the trick of renaming and wrapping additional tasks around an existing task would work fine, it's a bit coarse-grained.

This branch converts compile-theme into a multi-task and creates a new event to allow other files to "hook" in and add actions. It does not significantly change how the watch behavior operates, but that should be doable by adjusting the configuration after it is set by theme.js.

Pushing as a WIP for discussion.

### DIFF
--- a/tasks/theme.js
+++ b/tasks/theme.js
@@ -26,40 +26,41 @@ module.exports = function(grunt) {
    * }
    */
 
-  grunt.loadNpmTasks('grunt-contrib-compass');
-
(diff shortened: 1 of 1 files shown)


grayside:
This seems like going too far. After some consideration, dropping this. May resurrect it if frontend discussions lead us back here.

PR #104, opened 2015-01-07
Allow defaulting help descriptions in project Gruntconfig.json

grayside:
Help task requires configuration be loaded after bootstrap.js and before zzz_help.js is processed. This change allows projects to introduce custom task descriptions in Gruntfile.json so bootstrap.js picks it up.

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -6,9 +6,9 @@ module.exports = function(grunt) {
   });
 
   // Allow Gruntconfig.json to set default descriptions.
+  // tasks/zzz_help.js checks the root 'help' config namespace.
   if (config.help) {
     grunt.config('help', config.help);
-    grunt.config('config.help', '');
   }
 
   // Wrap Grunt's loadNpmTasks() function to change the current directory to
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -5,6 +5,12 @@ module.exports = function(grunt) {
     config: config
   });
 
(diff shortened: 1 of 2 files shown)


bootstrap.js
```diff
@@ -5,6 +5,12 @@ module.exports = function(grunt) {
     config: config
   });
 
+  // Allow Gruntconfig.json to set default descriptions.
+  if (config.help) {
+    grunt.config('help', config.help);
+    grunt.config('config.help', '');
```
arithmetric: @grayside, is this line necessary? I don't know of a use of `config.help` as a Grunt task.

arithmetric:
@grayside, could you add documentation on how this is used in Gruntconfig.json?

bootstrap.js
grayside: So Gruntconfig will get a 'help' key. When we load it into the grunt.config, it's located at grunt.config('config.help'). This code snippet is intended to move grunt.config('config.help') to grunt.config('help'). As an alternative, we could set up the help task to process config.help, but that seems super verbose.

arithmetric: If there's no functional value in removing the setting in 'config.help', then let's leave it there.

grayside:
Tests pass, cleanup of config.help removed, configuration documentation added.

arithmetric:
Thanks! :+1:

PR #106, opened 2015-01-07
Follow changes to grunt-drush.

IslandUsurper:
grunt-drush will no longer use this.files, as of nickpack/grunt-drush#11. Filenames for drush commands (such as make) should be added to the end of "args". While "cwd" is built-in option for files, it needs to be in an "options" object for task- and target-level configuration.


arithmetric:
Hi @IslandUsurper,

Thanks for the PR!

I tested this, and see that it doesn't work with the current grunt-drush v0.0.5, but works by using this patch-2 branch you refer:
https://github.com/chasingmaxwell/grunt-drush/tree/patch-2

So, we'll hold on merging this for now, and track this PR to grunt-drush.

Also, due to recent changes in grunt-drush, I've pinned its version to v0.0.3. So this change should include updating the pinned version to the version of grunt-drush that includes the above change.

IslandUsurper:
OK. grunt-drush 0.0.6 is up, and support for this.files has actually been restored, though it is no longer necessary.

IslandUsurper:
Hm. Should grunt-drush be pinned to 0.0.6 or ~0.0.6?

arithmetric:
Thanks for the update @IslandUsurper.

I'm leaning towards pinning it to 0.0.6 (and not allowing later versions). My practice has been to allow patch-level changes if the package has a minor release (~0.1.2 could float to 0.1.9 and minor-level changes if the package has a major release (^1.2.3 could float to 1.9.9). Since grunt-drush does not even have a minor release, the pattern would suggest that the patch-level is not allowed to float.

I'll review this PR later tonight and decide then whether to change the grunt-drush version.

(cc: @grayside )

PR #108, opened 2015-01-10
Adding test for Compass Sass compilation functionality

### LINKED ISSUE #69: Add test for compass functionality
Let's add some test coverage that compass is working, which incidentally should cover bundleInstall.

Doing so will require:

* A Gemfile to pull in compass dependencies.
* A Drupal theme-like structure that will allow testing config.rb & sass compilation, but does not otherwise need to be functional.
* Code to set up the above two things as part of the working copy for testing.
* A mocha to look for a generated CSS file.

We do not want these changes built directly into the examples folder, as examples should remain available as a stripped down template for starting projects, and SASS functionality should not be part of that.

arithmetric:
Also, adds `test/test_asset` structure to separate assets needed for testing from the base example. Adds `test/test_asset/src/modules/test.php` file to replace the file currently created by `create_working_copy.sh`.

Resolves #69.


test/test_assets/Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
```
grayside: By adding a Gemfile, we introduce a hard dependency on ruby, rubygems, and bundler in order to run tests. We should document this in README.md or EXTEND.md.

PR #110, opened 2015-01-12
Allow Gruntconfig to specify phpcs failures throw error.

grayside:
Simple override to specify that PHPCS failures should fail the build. Useful for strict projects. Waiting on Travis for actual testing.


grayside:
Also considered that `grunt.config('phpcs.options.ignoreExitCode', false)` may be simple enough to just document as an approach. No preference either way, but it seems like we lean a bit towards Gruntconfig carrying this kind of thing.

PR #115, opened 2015-01-16
Dynamic Concurrency [BC Break]

### LINKED ISSUE #103: Add dynamic concurrency level
Follow up from #78  and #94: We should add a universal concurrency setting that is applied to a number of tasks that can be run concurrently (drush make, Behat, etc.) and also identify a default value dynamically per @grayside's suggestion (based on grunt-concurent, see: https://github.com/sindresorhus/grunt-concurrent/blob/master/tasks/concurrent.js#L12).

grayside:
This change switches from Gruntconfig.json-based Drush make concurrency to something dynamic. It leverages the safe concurrency algorithm from the grunt-concurrent plugin and injects that into the assembly of the make parameters.

In case someone doesn't like the default, it also allows manually specifying `--concurrency=#` in the command-line invocation to override. Along the way I went ahead and added flag documentation to the README.

**BC Break: I couldn't find a graceful way to allow the Gruntconfig to retain override capability over this parameter, but given the sensitivity of this functionality across environments it struck me as probably not as useful as having the CLI option anyway.**

Couldn't easily run the new test, so letting Travis report on it.

Fixes #103

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # CHANGELOG
 
+## NEXT
+
+- Dynamically compute Drush Make concurrency based on system capability with a
+  new concurrency detection service.
+
+### Upgrade Notes
+
+- Remove --concurrency flag from your Gruntconfig. It will no logner be respected.
+
 ## v0.5.2 [2015/01/24]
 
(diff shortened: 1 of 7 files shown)


arithmetric:
@grayside, this looks good. Could you resolve the merge conflict?

PR #117, opened 2015-01-24
Changing bundle install default path options

arithmetric:
Changing bundle install default options to install gems to vendor/bundle/ and binaries to bin/.

### DIFF
--- a/tasks/bundler.js
+++ b/tasks/bundler.js
@@ -10,7 +10,7 @@ module.exports = function(grunt) {
 
   if (grunt.file.exists('Gemfile')) {
     grunt.config(['shell', 'bundle-install'], {
-      command: "bundle install --binstubs=gems/bin"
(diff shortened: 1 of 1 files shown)


grayside:
1. What happens if bundle install runs before composer? Both for the creation of the `bin/` and of `vendor/`?
2. Since we're doubling down on `repo:bin/`, do we know if composer or bundler configured this way disrupts custom `bin/` contents? That's we're I figure most projects should be putting their random scripts.

arithmetric:
Alternatively, should we put all binaries installed by Composer and Bundler into vendor/bin/.

grayside:
Are we concerned about a ruby vs. php name collision?

arithmetric:
@grayside I'm not very concerned about name collision between binaries installed by Composer and Bundler.

grayside:
I ended up pushing this as a new branch in case we changed our minds. If #123 is merged this can be closed.

PR #119, opened 2015-01-27
Fixing file name in the EXTEND.md documentation

arithmetric:
Fixing the EXTEND.md documentation for the correct file to add override JavaScript code.


grayside:
Looks good. The example right below is buggy, it overrides the configuration of all phplint. Will PR that later.

PR #120, opened 2015-01-28
Implement push instead of pull-based approach to help

### LINKED ISSUE #118: Eliminate load - order significance from help task
Current implementation pulls in configuration at time of plugin load. As a result any help configuration set after tasks/zzz_help is loaded does not work.

To support effective extensibility this limitation needs to be removed. The recently released gruntconfig.json option still creates a dependency on project config that hinders gdt-aware plugins from participating in help independently.

I've pushed a branch that creates an api to add config directly but it's only working when accessed from bootstrap.js.

grayside:
Fixed #118 

This branch removes most of the code from what was tasks/zzz_help.js and instead defines a small library module that provides a means for any other code loaded by grunt to push help content into the availabletasks plugin before the task is executed.

This also demonstrates a GDT helper module loader pattern so we don't need lots of requires everywhere. I think we may be able to do something clever like attaching gdt to grunt in bootstrap, but I thought to leave that notion alone for now.

Note that this still needs a documentation update, and completely discards both the pull-based mechanism we had before, as well as the more recent Gruntconfig-based help. I think it's reasonable to put back the Gruntconfig help if we want it, but would kind of like to just drop it as this is much more flexible and not that much more complicated for projects to implement.

### DIFF
--- a/lib/gdt.js
+++ b/lib/gdt.js
@@ -1,5 +1,0 @@
-module.exports = function(grunt) {
-  return {
-    help: require('./help')(grunt)
-  };
-}
--- a/lib/gdt.js
+++ b/lib/gdt.js
@@ -1,0 +1,5 @@
+module.exports = function(grunt) {
+  return {
+    help: require('./help')(grunt)
+  };
+}
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     "grunt-phpmd": "0.1.1",
     "grunt-shell": "1.1.1",
     "lodash": "2.4.1",
-    "require-directory": "^2.0.0",
     "time-grunt": "1.0.0"
   },
   "peerDependencies": {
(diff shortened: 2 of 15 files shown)


grayside:
Assigning to @arithmetric for methodology review, will add docs after we confirm this direction.

lib/gdt.js
```diff
@@ -0,0 +1,5 @@
+module.exports = function(grunt) {
+  return {
+    help: require('./help')(grunt)
+  };
+}
```
arithmetric: Let's hold off on the GDT abstraction until we have a larger library system.

package.json
```diff
@@ -23,6 +23,7 @@
     "grunt-phpmd": "0.1.1",
     "grunt-shell": "1.1.1",
     "lodash": "2.4.1",
+    "require-directory": "^2.0.0",
```
arithmetric: It doesn't seem like this is being used for this implementation. Is that right?

arithmetric:
@grayside, overall I like this approach. I'd prefer merging a simplified version, without the GDT abstraction.

PR #122, opened 2015-01-29
Task/minimize config buildpaths

grayside:
Moves the buildPaths config into implicit as part of the bootstrap.js

While I was in there, I also added a quick mkdir:init before analyze, as currently if you run analyze without having built the site it chokes on the lack of build/reports.


grayside:
I've updated the test/test_assets/Gruntconfig.json to strip out the buildPaths, as I'd thought we were still using the example/Gruntconfig.json for that.

For testing, how about we take a less e2e approach by loading a tailored Gruntconfig.json and verifying the config values we end up with after bootstrap?

Issue #102, opened 2015-01-05
Add option for dependency paths


arithmetric:
Add Gruntconfig.json option for the executable path for dependent programs/scripts, like drush, phpcs, phpmd, behat, etc.

By default, GDT should default to the path where the utilities are installed by Composer.

arithmetric:
See the discussion in #93.

arithmetric:
Note: #106 added handling for the path of drush, and added a default value to the Gruntconfig.json.

grayside:
Many of the commands used are now subject to this, we need an audit of the code to look for anything else and apply the needed changes. This is a very good first-time contributor issue as the approach is already laid out with some commands and it requires a tour of the codebase.

PR #123, opened 2015-02-07
Consolidate bundler and composer install paths and reclaim bin/ directory

grayside:
Takes another step on https://github.com/phase2/grunt-drupal-tasks/pull/117.

I've tested it backwards and forwards a bit, so it's looking pretty good. I seem to have a problem with grunt composer in my test environment, so let's see what Travis has to say.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # CHANGELOG
 
+## NEXT
+
+- Allow composer-installed binaries to be symlinked from the default directory
+  of `vendor/bin` to facilitate custom scripts in `bin/`.
+- Install Ruby libraries in `vendor/bundler` and binaries in `vendor/bin`.
+  `bundle exec` should continue to execute as normal, but see Upgrade Notes.
+
+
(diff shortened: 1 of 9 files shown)


grayside:
Looks like Travis can't find the drush instance at vendor/bin/drush.

arithmetric:
@grayside, I think the paths in test/test_assets/Gruntconfig.json need to be updated. See:
https://github.com/phase2/grunt-drupal-tasks/blob/feature/bundle_install_local-vendor_bin/test/test_assets/Gruntconfig.json

grayside:
Thanks! I forgot about that one. Looks like I have a couple PRs to clean up.

arithmetric:
This also changes the location of behat, which causes a break in behat tests. See the Travis script output:
```
stderr: 
/bin/sh: 1: ./bin/behat: not found
```

Currently the './bin/behat' path is hard-coded, so as a first step this hard-coded path can be updated. Ideally this path would be configurable like other dependencies, but that could be addressed in #102.

PR #131, opened 2015-03-19
Allow themes to take over their own grunt operations

grayside:
Different tells the story. Is this what we want?

When activated, can dispatch any task as `grunt theme:name:task` though it's a bit braindead on complex passthru. Tasks not wired into GDT can be added with that naming via a project - specific Gruntfile.js

@arithmetric @evanlovely


EvanLovely:
Love the idea! So what would be the command for running `grunt build` in the theme? 

PS I'm planning on changing the grunt command names to reflect the [spec document](https://docs.google.com/a/phase2technology.com/document/d/1IJIBFHcTpZRbb3RbYKMRPZ4bVDEnzLKujceP4hX3RoY/edit) you shared; so that command will become `grunt compile` eventually.

grayside:
We're not perfectly in-line with that spec here, a lot of evolution happened since it was updated. Or maybe devolution. :)

GDT's top-level "make the themes operational" task is `grunt compile-theme`.

To run `grunt build`, this change provides support for `grunt theme:serenade:build`. This could be tacked into grunt compile-theme via the [wrapping technique](https://github.com/phase2/grunt-drupal-tasks/blob/master/EXTEND.md#changing-the-build-process).

grayside:
As an alternative to extend it, I wonder if the ground has shifted enough that my [experiment with events](https://github.com/phase2/grunt-drupal-tasks/blob/feature/extend-theme-tasks/tasks/theme.js#L84) might be worthwhile.

grayside:
Per discussion, instead of dispatching individual pieces of functionality such as `grunt theme:serenade:compass`, we should attach `grunt theme:serenade:compile` to `grunt compile-theme`

tasks/theme.js
```diff
@@ -47,14 +51,15 @@ module.exports = function(grunt) {
           }, options)
         });
 
-        steps.push('compass:' + key);
+        task = theme.proxy ? 'theme:' + key + ':compass' : 'compass:' + key;
```
grayside: Check for proxy status above the compass attribute check. If so, add `grunt theme:key:compile` to the steps array. Skip the rest of the loop.

tasks/theme.js
```diff
@@ -20,6 +20,10 @@ module.exports = function(grunt) {
    *       "sourcemap": true
    *     }
    *   },
+   *   "legionaire": {
+   *     "path": "<%= config.srcPaths.drupal %>/themes/legionaire",
+   *     "proxy": true
```
arithmetric: Should we allow configuration of the grunt command invoked in the theme rather than mandate that they provide a specific command name?

grayside: For the task automatically attached to compile-theme? That makes sense. How about proxy property can instead be an object, with a default property that will override "compile"?

I hesitate to default to 'default' because that is often a watch task and task-not-found seems like a more obvious error state.

grayside:
I've made the changes discussed but have not had a chance to test.

CONFIG.md
```diff
@@ -275,6 +275,55 @@ directly to
 [grunt-contrib-compass](https://github.com/gruntjs/grunt-contrib-compass)
 for this theme.
 
+### Proxying Grunt Tasks
+
+Many themes come with their own Grunt integration. In those cases you can
+configure Grunt-Drupal-Tasks for awareness of that and remove the theme
+from most handling.
+
+In order to exclude your theme from handling, specify the `proxy` property in
```
arithmetric: To mirror the `compass` config option, let's name this property `grunt`.

CONFIG.md
```diff
@@ -275,6 +275,55 @@ directly to
...
+will be run inside your theme directory at the specified path.
+
+If you wish to customize the grunt task used, you can use a more complicated
+configuration:
+
+```
+"gladiator": {
+  "path": "<%= config.srcPaths.drupal %>/themes/gladiator",
+  "proxy": {
+    "default": "build",
```
arithmetric: Do we need the ability to have multiple proxy values here? Or could we use a simpler format:
```
"gladiator": {
  "path": "<%= config.srcPaths.drupal %>/themes/gladiator",
  "grunt": "build"
},
```

arithmetric:
@grayside I added two comments on the Gruntconfig format for this feature. Let me know your thoughts on them.

Based on the work integrating with Pattern Lab Starter, I'm ready to merge this soon.

CONFIG.md
grayside: I had actually considered going the other way, not mandating the theme use grunt in case there are gulp users.

CONFIG.md
grayside: Primary reason is to be able to extend to other tasks in future, e.g. dispatch validate for theme linting.

CONFIG.md
arithmetric: If we wanted to support gulp or a shell script, wouldn't we want to use a different property name, like `grunt` vs. `gulp` vs. `shell`?

CONFIG.md
arithmetric: :+1:

CONFIG.md
grayside: That seems like a potential source of confusion & complexity, since there are multiple non-default options that can individually trigger the "dispatch".

`themes.themeName.build = "grunt build"` or `themes.themeName.validate = "grunt eslint"`? If nothing is set it uses the GDT functionality. The downside of being an open shell command is that we lose the ability to pass along any other flags, such as if we wanted to pass along `--verbose`.

We could re-introduce some complexity later on if we see a need for something like that by detection of an object vs. a string:

`themes.invasion.build = { cmd: "grunt", args: "build" }`

arithmetric: On further consideration, I'm subscribing to the generic `proxy` approach. It seems like it would not be especially useful to differentiate if GDT is proxying to a Grunt- vs. Gulp- vs. shell-based theme script, since we'd invoke any of these scripts through a grunt-shell task. Given this, I'm learning towards including the command in the config structure (at least for the more advanced approach).

What do you mean by losing "the ability to pass along any other flags"? Do you mean flags that GDT would automatically apply or flags that could be configured in Gruntconfig.json? I'm not sure how I feel about automatically passing options through to the theme script, since it could be possible to include any desired options in the Gruntconfig.json option for the script.

What should the default task mappings be? (Or should we even have a simple default?) Some theme scripts may only have one appropriate mapping: GDT's `compile-theme` (also part of GDT's `default`) triggers a task in the theme. Others may have a corollary to `validate` or `test`, perhaps even `scaffold`.

How does this seem for the advanced configuration where you can map GDT tasks to theme script tasks, specify the command, task/argument, and command line option (optionally):

```
"legionaire": {
  "path": "<%= config.srcPaths.drupal %>/themes/legionaire",
  "proxy": {
    "compile-theme": "grunt compile --force",
    "validate": "grunt lint"
  }
}
```

tasks/theme.js
arithmetric: Note: This thread is superseded by the discussion above.

CONFIG.md
grayside: That makes sense to me. Strike the passthru, I can't think of a compelling use case.

The one downside I see is your config example is that proxy.compile-theme is a bit of a mouthful argument, especially given it's the most common mapping themes will use. Hence my use of "default".

arithmetric: I like brevity, but I think having consistency is important between the key name in proxy and the GDT task for which the proxy command is used.

grayside: Fair enough. So change is default => compile-theme. Anything else?

arithmetric: That change, and also removing the hard-coding of the "grunt" command in `command: 'grunt ' + task` so that this can be flexible for any kind of script.

grayside:
In pursuit of finalizing I've found some rough edges here.

* `grunt theme` errors
* `grunt theme:name` errors but should provide help
* `grunt theme:name:task` does not support targets or flags.

Going to polish some more with mild inspiration from npm scripts. 

That kind of enhancement raises the question whether themes without their own tooling should support ad hoc scripts from gdt, such as a `grunt theme:name:lsi` command that lists icon files for the theme from any directory of the project.

In that case, config.themes.name.proxy should be limited to a boolean and the command mapping moved to a scripts property. Following npm even closer we drop `proxy` and rely on magically named script keys.

grayside:
Summarizing the earlier comment, instead of `proxy`, allow themes the following configuration:

```js
{
    gladiator: {
        path: "<%= config.srcPaths.drupal %>/themes/gladiator",
        scripts: {
            'compile-theme': "grunt compile",
            validate: "grunt eslint",
            lsi: "ls -l icons/*.png" 
        }
    }
}
```

Like npm scripts, if a particular script key matches a GDT task, we "override" the theme's treatment to use the command. If the script does not match a GDT task, it's just an easy-to-execute task. We could also implement `grunt themes` => list all themes, `grunt themes:command` => run command against all themes which configure it, and `grunt themes:name:command` => run the command for the named theme.

I'll make it happen this week either way, but this approach is a bit more well-rounded and granular. Looking for input @arithmetric

arithmetric:
@grayside This config structure looks good to me. I also like the idea of the `themes` tasks, but that's less important for theme proxying.

grayside:
Changes as discussed recently:

1. `grunt themes` lists all configured themes
2. `grunt themes:name` lists all scripts configured for the specified theme
3. `grunt themes:name:script` uses grunt-shell to execute the configured value of script.
4. A test has been added to build.js to confirm a script runs successfully.
5. If scripts are configured with the keys "analyze", "validate", or "compile-theme" the script will be added to the tasks run by that GDT command. Note that this does not suppress compass if `"compass": true` has been configured for the theme.
6. I have enabled testing of tests/library.js, which has a couple tests in the unit style. Don't know why it wasn't running.
7. If the path is not specified for a theme, it will default to what you would expect. All it takes to add a theme to the system is `themes: { gladiator: {} }`, though this will only be useful for `grunt themes`.
8. Parts of the scripts system have been abstracted into lib because I anticipate reusing this pattern to bundle more general purpose operations into the Gruntconfig.

Things discussed but not present

* `grunt themes:all:script`. While I think there are use cases for it I decided not to keep going until the need is proven.
* `proxy` property to toggle whether overriding/dispatching/etc should be done. We're sticking to magic naming of script keys.
* A default command for theme compiling. Everything needs to be explicit. I think this might be a place where a generator could step to provide the baseline configuration for the theme.

PR #133, opened 2015-03-23
Switch to using composer to build drupal instead of drush make.

randallknutson:
With https://github.com/drupal-composer/drupal-project we could remove the dependency on drush make and use composer to install drupal instead. This would decrease the number of package managers being used to 3 (npm, composer, gem/bundler).

### DIFF
--- a/example/composer.json
+++ b/example/composer.json
@@ -1,15 +1,44 @@
 {
-    "name": "client/project",
-    "description": "{Project} drupal codebase for {client}.",
-    "require-dev": {
-        "behat/mink-zombie-driver": "~1.2",
-        "drupal/drupal-extension": "~3.0",
-        "drush/drush": "dev-master",
-        "phpmd/phpmd": "~2.1",
-        "squizlabs/php_codesniffer": "~1.5",
-        "drupal/coder": "7.2.3"
+  "name": "client/project",
+  "description": "{Project} drupal codebase for {client}.",
+  "repositories": [{
+    "type": "composer",
+    "url": "http://packagist.drupal-composer.org"
+  }],
+  "require": {
+    "roave/security-advisories": "dev-master",
+    "composer/installers": "~1.0",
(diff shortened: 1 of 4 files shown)


grayside:
This is really interesting. I think we'd want to retain support for drush make and enable this via config, but possibly default to composer after some live-fire testing.

It does commit you to forking workflows instead of patches , which may not be bad. I could see the nature of projects like progenitor shifting to emphasize fork maintenance.

randallknutson:
I added drush make back in as it was premature to remove it. Also, we aren't limited to forking. Patches are supported with https://github.com/netresearch/composer-patches-plugin. See: https://github.com/drupal-composer/drupal-project (Patches)

example/composer.json
```diff
@@ -1,15 +1,44 @@
...
+    "drupal/drupal-extension": "~3.0",
+    "drush/drush": "dev-master",
+    "phpmd/phpmd": "~2.1",
+    "squizlabs/php_codesniffer": "~1.5",
+    "drupal/coder": "7.2.3"
+  },
+  "config": {
+    "vendor-dir": "vendor"
+  },
+  "extra": {
```
grayside: Probably need two different composer.json files, as this one is too beefy for those using drush make still

tasks/composer.js
```diff
@@ -18,5 +19,31 @@ module.exports = function(grunt) {
       group: 'Dependency Management',
       description: 'Install dependencies defined in this project\'s composer.json file.'
     });
+
+    grunt.registerTask('composerinstall', 'Prepare the build directory and run "composer install"', function() {
```
grayside: Drushmake aside, we've standardized on hyphenating.

grayside:
I've done what code review can do. I need to try thus out before I can give a thumbs up. 

Doc updates are a blocker once the notion and code is reviewed.

arithmetric:
We'll continue to pursue this in the generator-gadget project.

PR #134, opened 2015-03-24
Create a Drupal-adaptive scaffold task. [WIP]

grayside:


arithmetric:
I've tested this with the default test package and also after dropping a Drupal 8 Drush makefile into test/test_assets/src/project.make, and found it to work with both the Drupal 7 and 8 sites.

grayside:
YES!

PR #141, opened 2015-03-28
Update test suite for Drupal 8

arithmetric:
Adding a Drush make file and updated Behat feature so that the test script builds Drupal 8 and runs Behat tests against it.

Should we also make this more explicit as another example for setting up GDT?


grayside:
I like the idea of some really basic behat tests as part of the example. Generic site validation seems like a good thing.

Not sure about dropping D7 tests while a majority of sites use it. I'd say unit tests around lib/drupal.js might be a middle ground but that file is so simple.

arithmetric:
I updated this PR to support testing with either Drupal 7 or Drupal 8. The Drupal core version is determined from an env var, and test assets and criteria can be specific to the core version.

.travis.yml
```diff
@@ -1,23 +1,27 @@
 language: php
 php:
   - 5.4
+env:
+  - GDT_DRUPAL_CORE="7" GDT_TEST_URL="http://127.0.0.1:8080/misc/drupal.js"
```
grayside: We already have GRUNT_DRUPAL_QUIET, we should probably align env variable prefixes. We could make GRUNT_DRUPAL_ deprecated and use GDT_ going forward.

test/create_working_copy.sh
```diff
@@ -11,16 +11,19 @@ if [ -d "test/working_copy/src/sites/default" ]; then chmod 755 test/working_cop
...
-cp -r example/* test/working_copy
+cp -r example/* $PATH_WORKING_COPY
 
 # Copy the test assets into working_copy
-cd test/test_assets
-for file in `find . -type f`; do mkdir -p ../working_copy/`dirname $file`; cp $file ../working_copy/$file; done;
+PATH_TEST_ASSETS="test/test_assets"
+if [[ $GDT_DRUPAL_CORE == "8" ]]; then PATH_TEST_ASSETS="test/test_assets_d8"; fi;
+cd $PATH_TEST_ASSETS
+for file in `find . -type f`; do mkdir -p $PATH_WORKING_COPY/`dirname $file`; cp $file $PATH_WORKING_COPY/$file; done;
```
grayside: More typical to use -exec? This seems fine.

test/test_assets_d8/Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
+gem 'sass', '~> 3.4.9'
+gem 'compass', '~> 1.0.1'
```
grayside: Adding a json library makes this a bit faster and is a best practice we've got documented: https://github.com/phase2/grunt-drupal-tasks/blob/master/CONFIG.md#setting-up-gem-dependencies

grayside:
My only review point that should hold this is the question about environment variable name prefixes. Otherwise :+1:

.travis.yml
arithmetric: I've updated the PR to change the GRUNT_DRUPAL_QUIET env var name to GDT_QUIET.

test/test_assets_d8/Gemfile
arithmetric: I've added the json gem to this file.

grayside:
Needs conflict resolution.

Also, Travis is listing contents of .git, do we need that?

arithmetric:
I've updated the PR to resolve the conflicts, and also moved the directory listing to the install block so it's not visible by default.

PR #142, opened 2015-03-28
Updating the loadNpmTasks wrapper function to limit its use to when the ...

### LINKED ISSUE #116: Trying to load other Grunt tasks fails because grunt-drupal-tasks monkeypatches grunt.loadNpmTasks()
I ran into a situation today where I tried to load another task from an npm module that put into my package.json file. Since grunt-drupal-tasks monkeypatches `grunt.loadNpmTasks()`, I could not load the tasks in question.

To work around this, it is necessary to either run `grunt.loadNpmTasks()` before the `require()` statement that loads grunt-drupal-tasks, or to run `grunt._loadNpmTasks()` to load your task if it comes after that require statement.

@mikecrittenden and I had discussed two possible solutions:
- Fall back and try to load tasks from the top-level node_modules directory after trying the initial load from the grunt-drupal-tasks node_modules directory.
- Detect errors in our monkeypatched `loadNpmTasks()` so that someone who hits this error will know what's going on.

arithmetric:
Updating the loadNpmTasks wrapper function to limit its use to when the module is found in GDT's node_modules directory.

Resolves #116.

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -1,5 +1,4 @@
-var path = require('path'),
-  _ = require('lodash');
+var _ = require('lodash');
 
 module.exports = function(grunt) {
   // Initialize global configuration variables.
@@ -23,7 +22,7 @@ module.exports = function(grunt) {
   // that are dependencies of Grunt Drupal Tasks.
   grunt._loadNpmTasks = grunt.loadNpmTasks;
   grunt.loadNpmTasks = function (mod) {
-    var internalMod = grunt.file.exists(path.join(__dirname, 'node_modules', mod));
+    var internalMod = grunt.file.exists(__dirname, 'node_modules', mod);
     if (internalMod) {
       var pathOrig = process.cwd();
       process.chdir(__dirname);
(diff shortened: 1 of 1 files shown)


smerrill:
LGTM :+1:

bootstrap.js
```diff
@@ -18,14 +19,19 @@ module.exports = function(grunt) {
...
 
-  // Wrap Grunt's loadNpmTasks() function to change the current directory to
-  // grunt-drupal-tasks, so that module dependencies of it are found.
+  // Wrap Grunt's loadNpmTasks() function to allow loading Grunt task modules
+  // that are dependencies of Grunt Drupal Tasks.
   grunt._loadNpmTasks = grunt.loadNpmTasks;
   grunt.loadNpmTasks = function (mod) {
-    var pathOrig = process.cwd();
-    process.chdir(__dirname);
+    var internalMod = grunt.file.exists(path.join(__dirname, 'node_modules', mod));
```
smerrill: It looks like `grunt.file.exists` will call `path.join` on all its arguments, so this one may be superfluous: https://github.com/gruntjs/grunt/blob/c6b534a7bc2e0a521f1fccf4963e58d58d3009e3/lib/grunt/file.js#L399

PR #145, opened 2015-03-29
Upgrades to 'grunt help'

### LINKED ISSUE #65: Document global CLI options in `grunt help`
It's not clear how to add help text for CLI options, like ``--timer``, in the `grunt help` display powered by grunt-available-tasks.

We need to explore how to document options like this in the help text displayed.

grayside:
This change adds greater complexity and reliance on the internals of the `grunt --help` system. However, that code seems pretty stable and changes are likely to be along the line of deprecating some of our changes here. I've come to the conclusion that the better experience is worth the code debt.

I copied the [output to a fancy gist](https://gist.github.com/grayside/06c412d4018da13d46af) to facilitate review.

Fixes #65

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ install:
   - npm install -g mocha
 script:
   - export PATH=$PATH:`pwd`/bin
+  - grunt help
   - grunt --timer
   - echo "Example fully built disk usage"; du -chs .
   - grunt drush:liteinstall --timer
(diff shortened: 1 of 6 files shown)


grayside:
Test failure is the result of github's present instability. Will trigger a re-test later.

arithmetric:
@grayside The Travis test is now fine, but when I tried `grunt help` locally, I found some errors and did not see the help display. See: https://gist.github.com/arithmetric/f6918bd7f4e99a1dd961

grayside:
I recall seeing that bug during development but it does not currently reproduce. Could you give me the output of --stack?

arithmetric:
I added `grunt help` to the Travis test script, and this exposed the same error I was seeing locally.

I found that this error could be avoided by removing the "gruntHelp.initTasks" routine from the help process queue.

PR #153, opened 2015-04-26
Add --notify option as a converse of --quiet

grayside:


grayside:
I haven't been able to confirm this is really working.

grayside:
Okay, got it working locally. Removing the WIP flag from the title.

PR #155, opened 2015-04-30
Allow tempbuild copy to copy dot-files.

bmcmurray:
Tells the files glob to include dot files in all directories. The tempbuild copy process did not automatically glob dotfiles included in downloaded modules when it copies to the build destination.


arithmetric:
Thanks @bmcmurray!

PR #157, opened 2015-05-16
Introducing Gadget

arithmetric:
Adding documentation on starting a project with Gadget.


grayside:
Nice! I think we should deep link to the new section from the main readme as part of a Gadget "ad".

PR #160, opened 2015-05-17
Add ESLint support with grunt-eslint

### LINKED ISSUE #42: Add support for CSS/SASS and JS lint

arithmetric:
Pulling #61 into a branch on this repo, and making a few updates.

The Drupal 8 .eslintrc should be added to the Gadget project template.

Resolves #42.


grayside:
I added a couple commits to the branch to help move this along: conflict resolution & .eslintignore. Since we're using .eslintrc we've got a generally usable eslint configuration that some things (e.g. IDEs) may automatically use, so we should make it friendly.

CONFIG.md
```diff
@@ -327,6 +327,29 @@ This is an example of the settings for the validate tasks:
...
 ```
 
+**eslint**: To enable eslint, set to `true` to use default options or an object
+with the following optional settings.
+
+**eslint.configFile**: The path to the eslint config file to use. If no value
+is specified, then `.eslintrc` in the project root is used.
+
+**eslint.ignoreError**: Set to `false` to trigger a Grunt error if validation
+fails, which will prevent subsequent tasks from executing.
```
grayside: I think the bigger deal is to fail builds on JS coding style errors.

Also, since we have a similar setting for phpcs, should we change this to a general validate.syntax.ignoreError setting? The granularity is nice but I can't really picture a case where I'd want to toggle different levels of strictness for different types of code.

example/.eslintignore
```diff
@@ -0,0 +1,19 @@
+##
```
grayside: I added this file.

example/.eslintrc
```diff
@@ -0,0 +1,29 @@
+{
```
grayside: This file doesn't match what I see in the [Drupal Code Repo](http://cgit.drupalcode.org/drupal/tree/.eslintrc), though maybe the current PR description means for it to be updated. My two questions are:

1. Does the D8 JS coding standard match what is expected of D7?
2. If (1) leads us to separate configs, I think we *should not* maintain our own copy of the D8 config. Instead we should do something clever like copy it from `build/html`.

tasks/quality.js
```diff
@@ -111,6 +114,25 @@ module.exports = function(grunt) {
...
+      eslintIgnoreError = eslintConfig.hasOwnProperty('ignoreError') ? eslintConfig.ignoreError : true,
+      eslintName = eslintIgnoreError ? 'force:eslint' : 'eslint';
+    grunt.config('eslint', {
+      options: {
+        configFile: eslintConfigFile
+      },
+      target: eslintTarget
+    });
+    validate.push(eslintName);
+    analyze.push(eslintName);
```
grayside: Right now everything in the analyze task is outputting to a file for Jenkins/etc to process. grunt-eslint/eslint does support the `checkstyle` format that phpcs uses.

example/.eslintrc
mdeltito: The existing .eslintrc file is a backport of the D8 rules for D7. I recently took a pass at adding support for both D7 and D8 rules, and encountered the same question/concern about trying to maintain a copy of the D8 rules file. 

Perhaps it would make the most sense to handle this in Gadget, and either copy in the D7 starter file OR download a copy of the D8 rules depending on the user's choice?

tasks/quality.js
arithmetric: @grayside I've updated the branch to differentiate the eslint:validate and eslint:analyze tasks.

example/.eslintrc
arithmetric: Agreed. This eslintrc is fine for the example. Let's improve the eslintrc that Gadget provides.

CONFIG.md
```diff
@@ -355,6 +355,13 @@ This is an example of the settings for the validate tasks:
 
 ```
 {
+  "eslint": {
+    "srcPaths": [
```
grayside: We've called this "dir" for PHPCS, should we try to make the property names parallel?

grayside: We've already got some issues like this, it should probably be it's own task to correct.

tasks/quality.js
```diff
@@ -49,8 +52,8 @@ module.exports = function(grunt) {
     var phpStandard = grunt.config('config.phpcs.standard')
       || 'vendor/drupal/coder/coder_sniffer/Drupal';
 
-    var ignoreError = grunt.config('config.phpcs.ignoreExitCode');
-    ignoreError = ignoreError === undefined ? true : ignoreError;
+    var ignoreError = grunt.config('config.validate.ignoreError');
```
grayside: We should probably preserve the existing config as functioning to avoid the immediate B/C break. Mark as deprecated and possibly output a warning?

arithmetric: @grayside Do you mean we should allow using either `config.validate.ignoreError` or `config.phpcs.ignoreExitCode`?

grayside: Yes, with the latter marked deprecated and pulled as part of #170.

arithmetric: Ok, preparing an update that supports either.

Issue #14, opened 2014-09-23
[META] Add performance testing


grayside:
We should add performance testing options. http://rupl.github.io/frontend-ops/#/7 points out a few solid options (scroll down).
- **yslow**: No API key needed, graded score.
- **phantomas**: Track performance trends.
- **grunt-perfbudget**: Lots of great data, and useful to specify performance budgets. http://timkadlec.com/2014/05/performance-budgeting-with-grunt/

I'm a little hazy on what I'm suggesting as far as tasks, maybe a grunt test:performance task run as part of the analyze task. The various performance test mechanisms can use a shared configuration of URLs to ping with the tests.

"Jenkins integration" for reporting on this may be tricky, I suspect we'll want to embed results in the job description unless we can dig up an output format Jenkins can interpret.

grayside:
The multi-env PR needs to be finished up to enable a lot of this. Once done, we should consider no-key pagespeed as an initial piece.

grayside:
Now that the multi-env PR has been merged, we can start to move forward with this. I think we will want to identify tools for three classes of performance analysis.

* Best practices for WPO. This is represented by tools such as Google Pagespeed, YSlow, and WebPageTest
* Benchmarking tests to prepare real numbers and confirm the site is faster than a project-specific, conservative limit.
* Investigative tools for when performance is found wanting, such as facilitating XHProf/Flamegraph generation.

grayside:
Closing old issues that I do not expect to be tackled in the context of this project.

PR #161, opened 2015-05-18
Add "grunt performance" to bring in Google Pagespeed.

grayside:
Officially `grunt performance`, but also accessible as `grunt pagespeed`, this command runs desktop and mobile pagespeed checks against each siteUrl. You may configure a `performance.samplePaths` property to override the default of testing the homepage alone, the intent is to use this configuration for all performance testing that needs a stock set of paths.

Because many performance tools rely on Web APIs or Crawlers, this branch introduces a new concept: "Web Readiness". See the new documentation for an explanation. Longer-term if we were to introduce tools that don't depend on external web access we can make grunt performance available for activating those tools, and switch the current warning mechanism to point out the availability of further tasks.

To see this in action you need to run `grunt performance --web-ready` or `GDT_WEB_READY=1 grunt performance`.

This is a "best practices" tool as described in #14.


grayside:
After further discussion, we want to focus on tools that can be run inside the firewall. The branch will stick around for reference, and might provide some further documentation later.

PR #166, opened 2015-06-04
Project Operations via the scripts mechanism

grayside:
This change adapts the mechanism used for the new Theme Scripts system to support simple project-level operations via simple Gruntconfig.json settings. These can be one-liners entirely contained in the configuration, or they could call out to shell scripts elsewhere in the repository.

If you configure some scripts and run `grunt help` you will see each script displayed in the Operations section.

The documentation has had a first pass, but I think it's getting confusing. Also, I think this needs a bit more test coverage, hence the WIP flag.

I've deliberately avoided setting any default scripts, but there's a lot of potential for gadget to generate a set of good starting ones by prompting for a profile name and maybe a couple other questions. I'll file some issues in the gadget queue based on this once we get it in.

Fixes #156, #98.

PR #178, opened 2015-06-17
Run Behat with process env vars

arithmetric:
Ensure that Behat is run with the process's environment variables in addition to ones added by the Behat task.

@peterschuelke reported an issue where Behat failed to run through Grunt Drupal Tasks with the command: `/bin/sh -c vendor/bin/behat -c ./test/behat.yml --tags ~@wip ././test/features/dashboard.feature` but could be run manually with: `vendor/bin/behat -c ./test/behat.yml --tags ~@wip ././test/features/dashboard.feature`. Another environment had an issue running Behat where it reported php not being found.

After applying this patch, Behat was able to run through Grunt Drupal Tasks using the 'sh -c' shell wrapper.

### DIFF
--- a/tasks/behat.js
+++ b/tasks/behat.js
@@ -61,9 +61,9 @@ module.exports = function(grunt) {
               maxProcesses: 5,
               bin: 'vendor/bin/behat',
               debug: true,
-              env: {
+              env: _.extend({}, process.env, {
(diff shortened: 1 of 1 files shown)


grayside:
Looks good, however is it worth considering whether this should be a bit of "magic" configuration as invented for https://github.com/phase2/grunt-drupal-tasks/pull/166/files. The main difference is you need to be more explicit about env vars of interest, but as a side benefit it provides a mechanism for these values to be set programmatically in project Gruntfile.js.

arithmetric:
@grayside That may be worth considering in a future version. Unless there are reasons to think this causes unexpected and bad consequences, we should get this in as a bug fix/usability enhancement and roll a patch release.

PR #179, opened 2015-06-23
Fixing error with 'grunt serve'

arithmetric:
Fixing error with 'grunt serve' due to drush:serve not being defined.

See: https://gist.github.com/arithmetric/89cbe01cf6063af8e1be


grayside:
That's strange, given I successfully tested it nested. In any case, this is fine.

PR #180, opened 2015-06-25
symlink libraries as part of scaffolding

spacepants:
This adds symlinking of libraries in `src/libraries` as part of the default scaffold task.

You can either manually include whatever libraries your code requires, or if you're using something fancy like [Bower](http://bower.io/), just point that to the `src/libraries` directory in your `.bowerrc`.

### DIFF
--- a/tasks/scaffold.js
+++ b/tasks/scaffold.js
@@ -20,9 +20,7 @@ module.exports = function(grunt) {
       grunt.loadNpmTasks('grunt-contrib-symlink');
 
       grunt.config(['symlink', 'libraries'], {
-        expand: true,
-        cwd: '<%= config.srcPaths.drupal %>/libraries',
-        src: ['*'],
+        src: '<%= config.srcPaths.drupal %>/libraries',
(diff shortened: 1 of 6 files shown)


spacepants:
Looks like composer just timed out when building dependencies. Everything looks good on my end.

Let me know if you need anything else to merge this in.

arithmetric:
Hi @spacepants,

Thanks for the contribution!

I agree that there should be a way to bundle libraries under src/. However, we often include libraries using a Drush makefile. If Drush make installs a library, then this symlink task fails, because the libraries directory already exists.

```
Running "symlink:libraries" (symlink) task
>> Destination build/html/sites/all/libraries already exists.
>> Created 0 symbolic links.
```

Could you update this task to symlink the individual directories in src/libraries instead of the entire directory? This would be similar to the task in Grunt Drupal Task's scaffold.js for the sites directories.

spacepants:
Thanks, @arithmetric. This should be all good to go.

arithmetric:
Thanks @spacepants. :+1:

PR #187, opened 2015-08-31
Add an install task

jhedstrom:
- Task works if no database is specified
- In manual testing, the database load only loads certain tables...

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -484,6 +484,27 @@ should be placed. Defaults to the package root. For Acquia set this to
 **packages.dest.devResources**: Specify where within the package directory the
 `projFiles` should be placed. Defaults to package root.
 
+### Install settings
+
+The `install` task will either install Drupal via the configured `profile`, or if
+a path to a database is configured, load this database:
+
+To specify a profile other than the `standard` profile:
+
+```json
+{
+  "serve": {
+    "profile": "openatrium"
+```
+
+to specify a database to load:
+
+```json
+{
+  "install": {
+    "db": "path/to/project.sql.gz"
+```
+
 ### Serve Settings
 
 The Serve task allows you to run Drupal using PHP's built-in webserver. This
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -493,7 +493,7 @@ To specify a profile other than the `standard` profile:
 
 ```json
 {
-  "serve": {
+  "project": {
     "profile": "openatrium"
 ```
 
@@ -501,9 +501,13 @@ to specify a database to load:
 
 ```json
 {
-  "install": {
+  "project": {
     "db": "path/to/project.sql.gz"
 ```
+
+**project.profile** The profile to use with `drush:site-install` internal task. Defaults to `standard` and may be overridden
+with `--profile` when the install task is run.
+**project.db** If specified, this database will be loaded instead of running a site installation.
 
 ### Serve Settings
 
@@ -515,10 +519,12 @@ terminal window.
 `grunt serve` will not install the Drupal site. Run with `grunt serve:demo` to
 skip starting up watch tasks.
 
-```
+```json
 {
+  "project": {
+    "profile": 'project_profile_name'
+  },
   "serve": {
-    "profile": 'project_profile_name',
     "port": 9043,
     "concurrent": [
       "watch-theme"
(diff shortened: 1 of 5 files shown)


tasks/install.js
```diff
@@ -0,0 +1,75 @@
...
+  grunt.config(['shell', 'loaddb'], {
+    command: 'gzip -dc <%= config.install.db %> | <%= config.install.dbConnection %>'
+  });
+
+  grunt.registerTask('install', 'Install Drupal with the configured profile, or from a database dump file if one exists.', function() {
+
+    // Check for a database file first.
+    var dbPath = grunt.config('config.install.db');
+    if (dbPath && grunt.file.exists(dbPath)) {
+      // @todo This doesn't work, the output of the sql-connect command is needed.
```
jhedstrom: I don't know how to dynamically capture the database credentials to use in the loaddb command below.

CONFIG.md
```diff
@@ -484,6 +484,27 @@ should be placed. Defaults to the package root. For Acquia set this to
...
+### Install settings
+
+The `install` task will either install Drupal via the configured `profile`, or if
+a path to a database is configured, load this database:
+
+To specify a profile other than the `standard` profile:
+
+```json
+{
+  "serve": {
```
jhedstrom: I wonder if we need to combine `serve` settings and the new `install`? I could just add the new db setting to the `serve` section if that is cleaner.

grayside: I'm thinking for B/C we should keep that for now, but maybe a top-level "project" key makes sense?

grayside:
Overall this looks good to me, but I want to give it another pass. It seems like our tests should switch to `grunt install`.

jhedstrom:
I took a bit of time looking into testing the sql dump reload part of this task, and there are some oddities with sqlite in terms of piping an sql dump that I haven't figured out how to address. We could do that in a follow-up, or alternatively just use mysql in the tests here.

grayside:
I think a from-scratch installation is more important to test than the DB load... or are we talking about doing both? If it's both I think restricting this functionality to MySQL is fine for now, we'll want to note that in the docs.

tasks/install.js
```diff
@@ -0,0 +1,81 @@
...
+      Drupal.loadDatabaseConnection(function(error) {
+        if (error) {
+          grunt.fail.fatal(error);
+          done();
+        }
+        grunt.task.run([
+          // Drop current database.
+          'drush:sql-drop',
+          // Load new database.
+          'shell:loaddb'
```
grayside: For a complete db load, I generally find the following next steps should be followed:

1. If using Features module: disable features rebuild on cache clear
2. Clear cache
3. Run updatedb to ensure database is synchronized with code.
4. If features module is used, run features-revert-all to ensure database is synchronized with code.

Could we somehow build that in here, or if all those steps seem unreasonable, could we trigger a grunt event so projects can tailor the process?

This seems fine as a follow-up.

tasks/serve.js
```diff
@@ -22,7 +22,11 @@ module.exports = function(grunt) {
   // If no path is configured for Drush, fallback to the system path.
   var cmd = {cmd: Drupal.drushPath()};
 
-  var profile = grunt.config('config.serve.profile') || 'standard';
+  // `config.serve.profile` is deprecated.
+  if (grunt.config('config.serve.profile')) {
+    grunt.log.warn('The `serve.profile` parameter is deprecated and will be removed in future versions. Use `install.profile` instead.');
```
grayside: Error message specified `install.profile`, config documentation uses `project.profile`. `project.profile` is better.

tasks/install.js
jhedstrom: Even after a site install I find an `fra` is usually needed. I can add those here.

grayside: Will need to check check for `fra` command or force through failures.

jhedstrom:
@grayside I made those changes. For the features revert I added `@todo` notes, because I didn't know offhand how to best determine if features is in use here or not.

grayside:
@jhedstrom discussed with @arithmetric , we concluded the best solution is to leave off the features-revert, and instead follow up with the project operations PR, adding in pre- and post- config options that can be used to add features revert or other tasks as needed.

Sorry to spin you around. I'll make a point of giving this a thorough manual testing tomorrow as I'm looking forward to using the functionality next week.

jhedstrom:
The features revert has been removed now.

PR #188, opened 2015-09-02
Feature/phpmd exclude config

spacepants:
This exposes phpmd's exclude to the gruntconfig, allowing per-project excludes.

### DIFF
--- a/tasks/quality.js
+++ b/tasks/quality.js
@@ -106,6 +106,9 @@ module.exports = function(grunt) {
 
   if (grunt.config.get('config.phpmd') != undefined) {
     var phpmdConfig = grunt.config.get('config.phpmd.configPath') || 'phpmd.xml';
+    var phpmdExclude = grunt.config.get('config.phpmd.exclude') || [
+      '<%= config.srcPaths.drupal %>/sites'
+    ];
     grunt.config('phpmd', {
       custom: {
         dir: '<%= config.srcPaths.drupal %>/'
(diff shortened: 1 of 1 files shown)


grayside:
Hi @spacepants, thanks for the PR but we've decided not to go this route.

grunt-drupal-tasks tackles so much, there's a danger of closely wrapping all the grunt plugins with our own configuration plugin layer, which is not our goal. Instead, we want to aim at a bit less managed flexibility with the configuration we manage and leave it as a developer exercise to implement a line such as the following in their project gruntfile.js:

```js
grunt.config('phpmd.options.exclude', grunt.config('config.custom.phpmd.exclude'));
```

or for a more sophisticated take, wherein we [preserve the existing configuration](http://gruntjs.com/api/grunt.config#grunt.config.getraw) and customize without breaking compatibility with other grunt-drupal-tasks configurations:

```js
var options = grunt.config.getRaw('phpmd.options.exclude');
grunt.config('phpmd.options.exclude', options + ', <%= config.srcPaths.drupal %>/libraries');
```

"Outlier configuration" such as this adds complexity and insulates grunt-drupal-tasks users from the underlying grunt plugins. I certainly appreciate how [phplint.dir](https://github.com/phase2/grunt-drupal-tasks/blob/master/tasks/quality.js#L43) and [eslint.dir](https://github.com/phase2/grunt-drupal-tasks/blob/master/tasks/quality.js#L127) settings create the impression we want to facilitate more configuration, but in those cases we are surfacing extremely common customizations that are more embedded in how grunt-drupal-tasks is handling consistency across targets.

PR #189, opened 2015-09-02
Add commands to README

mikecrittenden:
Not sure if you think this is useful or just more to maintain. I was creating it anyway for docs for MLS so I figured I'd PR it.

The thinking is that it's helpful for people to be able to see what GDT provides without having to download it and set it up first. Like, maybe someone wouldn't be interested until he/she saw that "grunt package" exists or something.

Anyways, feel free to ignore, won't hurt my feelings, but here it is!

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,62 @@ to view documentation tailored for your project.
 
 To build your Drupal site, run `grunt`.
 
+### Commands
+
+The following is a list of base commands and their options which you can expect to find on any **grunt-drupal-tasks** based project.
+
+#### Build Process
+
+* `grunt`
+  * The default build process, which includes verifying dependencies and assembling the build directory. 
+* `grunt scaffold`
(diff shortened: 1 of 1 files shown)


grayside:
Thanks @mikecrittenden.

We're talking through the maintenance overhead, and realized there might be a few things in `grunt help` that need to be fixed or removed from that listing. For example, `grunt drushmake` doesn't result in a usable codebase, so should we only list drush make as a step of the default grunt task, or does it need to be followed up with the `grunt scaffold` task automatically? Good questions as we get closer to a 1.0.0 roadmap.

Definitely hear you that we need to do better to describe the features in the README.

mikecrittenden:
Yup, that's all it comes down to - putting features on display. Again, feel 100% free closing this if you'd like to go about that another way.

grayside:
I think we've decided not to pull in the commands in this way. However, I am working on static site generation, maybe if that comes together we'll find a way to back the actual help output in?

PR #192, opened 2015-09-21
Prevents phpcs tasks being added if no files

### LINKED ISSUE #174: The phpcs:validate task never ends if there are no files
To reproduce:

```
yo gadget
grunt
```

It eventually (many minutes):

```
Running "phpcs:validate" (phpcs) task
Starting phpcs (target: validate) in src/**/*.css src/**/*.php src/**/*.module src/**/*.inc src/**/*.install src/**/*.profile !src/sites/** !src/**/*.box.inc !src/**/*.features.*inc !src/**/*.pages_default.inc !src/**/*.panelizer.inc !src/**/*.strongarm.inc
ERROR: You must supply at least one file or directory to process.

Usage: phpcs [-nwlsaepvi] [-d key[=value]]
    [--report=<report>] [--report-file=<reportfile>] [--report-<report>=<reportfile>] ...
    [--report-width=<reportWidth>] [--generator=<generator>] [--tab-width=<tabWidth>]
    [--severity=<severity>] [--error-severity=<severity>] [--warning-severity=<severity>]
    [--runtime-set key value] [--config-set key value] [--config-delete key] [--config-show]
    [--standard=<standard>] [--sniffs=<sniffs>] [--encoding=<encoding>]
    [--extensions=<extensions>] [--ignore=<patterns>] <file> ...
                      Set runtime value (see --config-set)
        -n            Do not print warnings (shortcut for --warning-severity=0)
        -w            Print both warnings and errors (on by default)
        -l            Local directory only, no recursion
        -s            Show sniff codes in all reports
        -a            Run interactively
        -e            Explain a standard by showing the sniffs it includes
        -p            Show progress of the run
        -v[v][v]      Print verbose output
        -i            Show a list of installed coding standards
        -d            Set the [key] php.ini value to [value] or [true] if value is omitted
        --help        Print this help message
        --version     Print version information
        <file>        One or more files and/or directories to check
        <extensions>  A comma separated list of file extensions to check
                      (only valid if checking a directory)
        <patterns>    A comma separated list of patterns to ignore files and directories
        <encoding>    The encoding of the files being checked (default is iso-8859-1)
        <sniffs>      A comma separated list of sniff codes to limit the check to
                      (all sniffs must be part of the specified standard)
        <severity>    The minimum severity required to display an error or warning
        <standard>    The name or path of the coding standard to use
        <tabWidth>    The number of spaces each tab represents
        <generator>   The name of a doc generator to use
                      (forces doc generation instead of checking)
        <report>      Print either the "full", "xml", "checkstyle", "csv", "json"
                      "emacs", "source", "summary", "svnblame", "gitblame", "hgblame" or
                      "notifysend" report
                      (the "full" report is printed by default)
        <reportfile>  Write the report to the specified file path
        <reportWidth> How many columns wide screen reports should be printed
```

EvanLovely:
This just expands the file globbing done on `phpcs` and sees if there's not 0 files before setting the config and adding it to the validate and analyze task. The diff is best [viewed with no whitespace](https://github.com/phase2/grunt-drupal-tasks/pull/192/files?w=0) as it's really just wrapping the code in `if (grunt.file.expand(phpcs).length !== 0) {}`. This should resolve #174

### DIFF
--- a/tasks/quality.js
+++ b/tasks/quality.js
@@ -46,8 +46,8 @@ module.exports = function(grunt) {
 
   if (grunt.config.get('config.phpcs') != undefined) {
     var phpcs = grunt.config.get('config.phpcs.dir') || [
-      '<%= config.srcPaths.drupal %>/**/*.css'
-    ].concat(defaultPatterns);
+        '<%= config.srcPaths.drupal %>/**/*.css'
+      ].concat(defaultPatterns);
 
     var phpStandard = grunt.config('config.phpcs.standard')
       || 'vendor/drupal/coder/coder_sniffer/Drupal';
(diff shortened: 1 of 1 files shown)


grayside:
It would be better to fix this upstream, but since the maintainer is on a low-velocity schedule it makes sense to get something in here. If we find the extra effort of double-computing the glob to be trouble, we can add a fix to https://github.com/grayside/grunt-phpcs.

PR #193, opened 2015-09-21
adding ability to skip `grunt:validate` step to default task

EvanLovely:
Run `grunt --no-validate` to skip `grunt:validate` in default `grunt` task. Helpful for troubleshooting as it skips the step that takes the longest or can potentially fail the whole build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ These flags are not yet documented as part of the `grunt help` command.
 * `--notify`: Request a desktop notification despite timing or environment settings.
 * `--timer`: Output execution time profile at the end of the task run.
 * `--concurrency`: Override the dynamic concurrency by Drush Make.
+* `--no-validate`: Skip `grunt:validate` in the default `grunt` task.
 
 ### Environment Options
 
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -36,6 +36,7 @@ module.exports = function(grunt) {
   if (grunt.task.exists('bundle-install')) {
     tasksDefault.push('bundle-install');
   }
+  // If the "--no-validate" option is given, skip adding "validate" to task array.
   if (!grunt.option('no-validate')) {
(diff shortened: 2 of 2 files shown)


bootstrap.js
```diff
@@ -28,17 +28,19 @@ module.exports = function(grunt) {
...
   }
   if (grunt.task.exists('bundle-install')) {
-    tasksDefault.unshift('bundle-install');
+    tasksDefault.push('bundle-install');
   }
+  if (!grunt.option('no-validate')) {
+    tasksDefault.push('validate');
+  }
+  tasksDefault.push('newer:drushmake:default');
+  tasksDefault.push('scaffold');
```
EvanLovely: The order of the tasks are still the same.

README.md
```diff
@@ -69,6 +69,7 @@ These flags are not yet documented as part of the `grunt help` command.
 * `--notify`: Request a desktop notification despite timing or environment settings.
 * `--timer`: Output execution time profile at the end of the task run.
 * `--concurrency`: Override the dynamic concurrency by Drush Make.
+* `--no-validate`: Skip `grunt:validate` in the default `grunt` task.
```
grayside: Needs to be added to tasks/help.js,

grayside:
Needs conflict resolution, and a merge from master should also fix the tests.

I was initially unsure about adding this, but I've since come around on a few points:

1. validate is often where build-blocking bugs get introduced
2. anyone wanting to build a clean CD pipeline will probably like the option to run validate as a separate task from the main build, or possibly skip the validate and use the analyze task results
3. The new git hooks feature ensures some minimal validation run is part of the system for local development where this option is abused.

@arithmetric what do you think?

arithmetric:
@grayside Those are good justifications for this option. I'm ok with a merge once this is added to help.js as you noted.

PR #194, opened 2015-09-22
Exclude theme from validate if it supplies it's own validate

EvanLovely:
If the theme has a validate task of it's own set in the theme scripts, then let's exclude it from GDT's `grunt:validate`.


jhedstrom:
Manual testing confirmed that this works.

Issue #24, opened 2014-10-18
Add support for deployment tooling


grayside:
When writing the package task, the intent was to support a swappable deployment mechanism. Our two base cases being a fairly straightforward Capistrano rsync method and a pantheon/acquia friendly  docroot commit to ops repo.

Nationally, these would be driven from grunt-shell tasks.

Also worth considering [ShipIt](http://www.therightcode.net/deploy-nodejs-application-using-shipit/)

grayside:
A separate release repository makes lots of sense, with possible value in a grunt plugin like https://www.npmjs.com/package/grunt-git-deploy. That would allow us to have some unification across Acquia, Pantheon, and other hosting environments.

However, we don't want that to be the only means GDT can support deployment, though it would be easy enough for a separate project to implement some other mechanism and override anything we might do with different behavior.

nahuelthc:
@grayside: could you explain more how you do it? I kind of put that content on files, but grunt don't put any think in that dir. I'm old php wolf, node and grunt still new territory for me.

nahuelthc:
I'm deploying to platform.sh, that is a lot like pantheon. I made a recipe with sh while we found a solution in js.
1. Made a new directory build/platform
2. Started a git repoctory there
2. Use this platform_deploy.sh in proyect root:
#!/bin/bash
# clean last run
rm -r build/platform/*
rm -r build/platform/.htaccess
rm -r build/platform/.DS_Store
# create package
grunt package
# use package.tgz
tar zxf build/packages/package.tgz -C build/platform docroot
mv build/platform/docroot/* build/platform
mv build/platform/docroot/.[!.]* build/platform
rm -r build/platform/docroot
# git
cd build/platform
git add *
git commit -m "auto commit"
git push origin master


Hope its help somebody else!

PR #195, opened 2015-10-23
Initial swing and proof of concept for acquia packaging.

mpgeek:
This is a work in progress. Issuing a PR for visibility and review. At this point i'm wondering if this is the right architecture for release packaging. Most importantly, packaging does a little disassembly of the `html` build (as the original `package` command does), and splits it out into subdirectories. This structure is inspired by an older clinet project that is directly deployed to Acquia as-is.

Another question here is how the packaging should be invoked. Right now it's

```bash
grunt package --target=acquia
```

I thought it might make sense to have the `:compress` modifier work regardless of vendor target, but no work towards that was done. Compressing still works with the default command, i.e. no `--target` specifier.

The code style definitely needs work, as this is really POC-ey. Further, no work towards the recieving repository has been considered. The idea there is that should be configured by the derivative project.

Finally, some thought might want to be thrown at how this will work with gadget. Does it make any sense to move the logic in its entirety to derivative project setup, rather than accounting for vendors on this side?

### DIFF
--- a/example/Gruntconfig.json
+++ b/example/Gruntconfig.json
@@ -1,6 +1,6 @@
 {
   "srcPaths": {
-    "make": "src/*.make",
+    "make": "src/project.make",
     "drupal": "src"
   },
   "domain": "project.vm",
@@ -9,7 +9,7 @@
   },
   "packages": {
     "srcFiles": ["!sites/*/files/**", "!xmlrpc.php", "!modules/php/*"],
-    "projFiles": ["README*", "bin/**", "hooks/**"]
+    "projFiles": ["README*", "bin/**", "hooks/**", "src/*.make"]
   },
   "phpcs": {
     "path": "vendor/bin/phpcs"
--- a/example/Gruntconfig.json
+++ b/example/Gruntconfig.json
@@ -5,8 +5,20 @@
   },
   "domain": "project.vm",
   "packages": {
-    "srcFiles": ["!sites/*/files/**", "!xmlrpc.php", "!modules/php/*"],
-    "projFiles": ["README*", "bin/**"]
+    "default": {
+      "srcFiles": ["!sites/*/files/**", "!xmlrpc.php", "!modules/php/*"],
+      "projFiles": ["README*", "bin/**"]
+    },
+    "acquia": {
+      "srcFiles": ["**", ".**", "!sites/default/**", "!sites/*/files/**", "!??/**", "!sites/all/modules/dev/**", "!sites/*/settings.php", "!sites/*/settings.db.inc", "!sites/*/settings.local.inc", "!.htaccess", "!.gitignore", "!.gitkeep"],
+      "sharedFiles": ["**", "!files/**", "!default.settings.php", "settings.php", "!settings.local.inc"],
+      "projFiles": ["README*", "bin/**"],
+      "root": "build/packages/acquia",
+      "paths": {
+        "docroot": "build/packages/acquia/docroot",
+        "shared": "build/packages/acquia/shared"
+      }
+    }
   },
   "phpcs": {
     "path": "vendor/bin/phpcs"
@@ -22,3 +34,4 @@
   },
   "eslint": true
 }
+
--- a/example/Gruntconfig.json
+++ b/example/Gruntconfig.json
@@ -1,24 +1,19 @@
 {
   "srcPaths": {
     "make": "src/project.make",
-    "drupal": "src"
+    "drupal": "src",
+    "devroot": "devroot"
   },
   "domain": "project.vm",
-  "packages": {
-    "default": {
-      "srcFiles": ["!sites/*/files/**", "!xmlrpc.php", "!modules/php/*"],
-      "projFiles": ["README*", "bin/**"]
-    },
-    "acquia": {
-      "srcFiles": ["**", ".**", "!sites/default/**", "!sites/*/files/**", "!??/**", "!sites/all/modules/dev/**", "!sites/*/settings.php", "!sites/*/settings.db.inc", "!sites/*/settings.local.inc", "!.htaccess", "!.gitignore", "!.gitkeep"],
-      "sharedFiles": ["**", "!files/**", "!default.settings.php", "settings.php", "!settings.local.inc"],
-      "projFiles": ["README*", "bin/**"],
-      "root": "build/packages/acquia",
-      "paths": {
-        "docroot": "build/packages/acquia/docroot",
-        "shared": "build/packages/acquia/shared"
-      }
+  "buildPaths": {
+    "package": {
+      "default": "build/packages/package",
+      "acquia": "build/packages/acquia"
     }
+  },
+  "packages": {
+    "srcFiles": ["!sites/*/files/**", "!xmlrpc.php", "!modules/php/*"],
+    "projFiles": ["README*", "bin/**"]
   },
   "phpcs": {
     "path": "vendor/bin/phpcs"
@@ -34,4 +29,3 @@
   },
   "eslint": true
 }
-
--- a/example/Gruntconfig.json
+++ b/example/Gruntconfig.json
@@ -1,19 +1,15 @@
 {
   "srcPaths": {
-    "make": "src/project.make",
-    "drupal": "src",
-    "devroot": "devroot"
+    "make": "src/*.make",
+    "drupal": "src"
   },
   "domain": "project.vm",
   "buildPaths": {
-    "package": {
-      "default": "build/packages/package",
-      "acquia": "build/packages/acquia"
-    }
+    "packages": "build/packages"
   },
   "packages": {
     "srcFiles": ["!sites/*/files/**", "!xmlrpc.php", "!modules/php/*"],
-    "projFiles": ["README*", "bin/**"]
+    "projFiles": ["README*", "bin/**", "hooks/**"]
   },
   "phpcs": {
     "path": "vendor/bin/phpcs"
--- a/tasks/package.js
+++ b/tasks/package.js
@@ -40,12 +40,6 @@ module.exports = function(grunt) {
           src: projFiles,
           dest: destPath + (grunt.config.get('config.packages.dest.devResources') || ''),
           dot: true
-        },
-        {
-          expand: true,
-          src: grunt.config.get('config.srcPaths.make'),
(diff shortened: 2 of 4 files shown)


mpgeek:
Related: https://github.com/phase2/grunt-drupal-tasks/issues/24

grayside:
Hey @mpgeek , thanks for getting this moving!

I'll in-line some nitpicks, but first a higher-level review.

### Concept Review

1. I like the notion to be able to compress any package target. Certainly it's not always necessary but if you want to keep a compressed build artifact somewhere it could be useful.
2. You've definitely got the idea on shaping packages to purpose.
3. A complete fix for #24 also needs pushing the package to the remote repo, that doesn't have to be part of this PR for it to go forward.

### Discussion Topic

Does a single project need multiple packaging profiles?

### Interface

I think `--target` seems fine to me. As an alternative, we could go with a structure like `grunt package:<target>[:compress]`. That might be faster to type but it seems less clear to me, so ++ to what you have.

example/Gruntconfig.json
```diff
@@ -5,8 +5,20 @@
...
-    "projFiles": ["README*", "bin/**"]
+    "default": {
+      "srcFiles": ["!sites/*/files/**", "!xmlrpc.php", "!modules/php/*"],
+      "projFiles": ["README*", "bin/**"]
+    },
+    "acquia": {
+      "srcFiles": ["**", ".**", "!sites/default/**", "!sites/*/files/**", "!??/**", "!sites/all/modules/dev/**", "!sites/*/settings.php", "!sites/*/settings.db.inc", "!sites/*/settings.local.inc", "!.htaccess", "!.gitignore", "!.gitkeep"],
+      "sharedFiles": ["**", "!files/**", "!default.settings.php", "settings.php", "!settings.local.inc"],
+      "projFiles": ["README*", "bin/**"],
+      "root": "build/packages/acquia",
```
grayside: I think "root" should be zero-config by appending the package "profile" key as a new path layer to the buildPaths.package config.

example/Gruntconfig.json
```diff
@@ -5,8 +5,20 @@
...
+      "paths": {
+        "docroot": "build/packages/acquia/docroot",
+        "shared": "build/packages/acquia/shared"
+      }
```
grayside: Look at https://github.com/phase2/grunt-drupal-tasks/blob/master/CONFIG.md#package-settings to see the package.dest settings.

What is the role of the shared directory?

tasks/package.js
```diff
@@ -71,4 +146,15 @@ module.exports = function(grunt) {
...
   });
+
+  function isValidTarget(target) {
+    if (target.length) {
+      var validTargets = ['acquia'];
+      return (validTargets.indexOf(target) > -1);
+    }
+    else {
+      return false;
+    }
```
grayside: I don't think we need to whitelist targets. Let people configure whichever keys they want to build out a target/profile. We should document any "magic keys" such as `acquia` that provide special sauce.

tasks/package.js
```diff
@@ -13,35 +13,110 @@ module.exports = function(grunt) {
...
+
+    // Build the specified vendor target, or the default.
+    switch (validTarget) {
+      case 'acquia':
+        /**
+         * Acquia release build tasks.
+         */
+        grunt.config.merge({
+          copy: { 'acquia_docroot': {
+            files: [
```
grayside: This is slightly less performant than: `grunt.config.set('copy.acquia_docroot', {...});`. grunt.config.merge is best for interlacing configuration rather than appending new sub-trees.

tasks/package.js
```diff
@@ -13,35 +13,110 @@ module.exports = function(grunt) {
...
+    }
+    else {
+      grunt.config('config.buildTarget', 'default');
+    }
+    grunt.log.writeln('Building package for ' + grunt.config.get('config.buildTarget'));
+
+    var validTarget = grunt.config.get('config.buildTarget');
+
+    // Build the specified vendor target, or the default.
+    switch (validTarget) {
```
grayside: Why do we need to custom wrangle Acquia? The only significant difference from the default expectations is the shared directory. As I asked in another comment, not sure the role of that directory but we can follow the pattern in place for dest.docroot and dest.devResources and have a generic loop for all things in package.dest to catch and copy them.

example/Gruntconfig.json
mpgeek: I followed the setup of the [Vista Campus repo](https://bitbucket.org/phase2tech/vista-campus/), where the `shared` directory is outside of docroot. @jhedstrom set that one up and can probably speak more to the architecture. I think this also speaks to the need to standardize. Since the only requirement for Acquia is that `docroot` is a root folder in the repo, builds might have needs for resources outside of that similar to Vista Campus. Mostly this was an exercise in seeing if this type of segmented structure could work. If we have a preferred repo architecture, we should document it and we can build that in.

example/Gruntconfig.json
mpgeek: Noted for next round.

tasks/package.js
mpgeek: Agreed. The wrangling was in the interest of supporting the segmented repo structure mentioned there. More habit here in validating user input than anything else. If the standard is "make acquia like `build/html`, then this is all cruft.

tasks/package.js
mpgeek: Noted for next round.

tasks/package.js
mpgeek: Noted for next round.

mpgeek:
It seems like once a project is set up for a specific vendor and build architecture, it would be set in stone unless some kind of hosting change took place. It might make sense to alias the package command to the specified build target, or even to make it hardcoded config somehow in that the specification is committable configuration. @jhedstrom's input on this initial effort eluded to that, but i didn't go that route on this initial swing. I like that idea better, i think.

example/Gruntconfig.json
grayside: I think that might align with what I meant by the package.dest.devResources option, will have to see what J has to say.

mpgeek:
This round was a significant simplification. Out of the box, `package` and `package:compress` work as they normally would without a bunch of extra cruft. If you want to make your package an Acquia package, you add two things to your projects `Gruntconfig.json`:

1. add `"packageTarget": "acquia"` in `"buildPaths"`
2. add `"dest": { "docroot": "/docroot" }` in `"packages"`

Under that scheme both making a release for Acquia, and a compressed version now work as well. I figure some naming or logic tightening could be in order, was hoping for some review here.

grayside:
1. What is the difference between devroot and the existing `packages.dest.devResources` option?
2. packageTarget as described doesn't make sense to fit in buildPath, but more significantly, why would you need to define multiple "packageTargets" in the system if it can only select one?

Maybe we should allow selection of a packaging "mode" via that --target parameter concept from the OP, and use default if none is expected. This assumes we want multiple packaging structures for a project as well--I haven't thought that through yet, but in general I would want to have an Acquia project's non-Acquia environments be as close a mirror as possible anyway.

mpgeek:
The `devroot` is where non-docroot resources live that are placed into the root of the package which make use of the `packages.dest.devResources` value for destination path. In this case it's the default value of `/`. In other words everything in `devroot` (sibling to `src`) ends up where `packages.dest.devResources` tells it to go.

Towards your second point, WW did exactly that: make the source repo more like the release repository, I guess the goal for this update was to leave the default working as it is, and make you work a little bit by adding a few things to turn the packager into an Acquia one.

But the discussion point of whether or not a project should be able to produce different packages with a `--target` parameter is valid. In the current state, obviously, you make a decision and commit it to your local configuration.

Its starting to seem like we have two choices here, one that is really just documentation (how to set up your local config to build acquia-compliant packages), and the other is allowing flexibility by taking input on the packaging decision (and having the acquia support baked in). This second option is more flexible and less work for the local project, but could easily be viewed as cruft when it can be done entirely in local configuration.

I think we need to decide if we want to go the minimal, mostly-documentation route, or the flexible route. I wonder if there is an intermediate stepping stone here to move forward with.

mpgeek:
Updated to be a blend of baked-in, general support, and documentation. Added help to the CONFIG.md for using the general packager as an acquia one. One item of note is that we don't depend on a specific package name anymore, and anything can either be passed in, or set in Gruntconfig. I left the example/Gruntconfig mostly alone so that default, OOTB packaging just works and added the naming support to package.js.

I think this is much simpler and addresses the points that Acquia is just a special use case for the packager, and that we can leave some work to the client project to set it up.

example/Gruntconfig.json
```diff
@@ -1,12 +1,15 @@
 {
   "srcPaths": {
-    "make": "src/project.make",
+    "make": "src/*.make",
```
jhedstrom: I think this needs to stay as `src/project.make`. We should add `src/*.make` to the `projFiles` example so those get packaged up.

mpgeek: Updated.

grayside:
I like how lean the change is getting. package target strikes me as a bit odd. What's the case to use it in Gruntconfig.json where you wouldn't just configure it as part of the main buildPaths.package option?

Maybe it should just be a be a CLI flag, and if so should it be relative or absolute?

Really close now, but there merge conflicts to to be resolved.

tasks/package.js
```diff
@@ -17,8 +17,14 @@ module.exports = function(grunt) {
       srcFiles = ['**', '!**/.gitkeep'].concat((config && config.srcFiles && config.srcFiles.length) ? config.srcFiles : '**'),
       projFiles = (config && config.projFiles && config.projFiles.length) ? config.projFiles : [];
 
-    var destPath = grunt.config.get('config.buildPaths.package') + '/package';
-    var tasks = []
+    // Look for a package target spec, build destination path.
+    var packageTarget = grunt.option('target') || process.env.GRUNT_PACKAGE_TARGET;
```
grayside: Revisiting this now, as I think the package stack will get some attention soon. The environment variable isn't a bad fallback, but each project is likely to have a single target needed ubiquitously. I think Gruntconfig.json overridden by a command-line flag will have much more use.

```
var packageTarget = grunt.option('target') || grunt.config.get('config.package.target');
```

In looking at the UI, I also wonder if package --name or --path might be a more intuitive naming convention.

grayside:
I understand that much of this was merged as part of a followup.

PR #198, opened 2015-10-25
Fix test failures on PHP 5.5+ for Drupal 8

grayside:
Did not look into why these combinations are taking more time, but the mocha default timeout of 2 seconds is being fairly routinely reached. I have bumped the limit to 10 seconds here and observed tests passing.

@arithmetric @jhedstrom


jhedstrom:
Great to see these all green, even on PHP 7. :+1:

PR #205, opened 2015-11-11
Use the pv command if available for loading the database.

jhedstrom:
This successfully uses the `pv` command, but grunt, even with the `-v` option is swallowing the output. Posted here in case anybody knows how to get that output through.


grayside:
Stale for almost 2 years without solution.

PR #207, opened 2015-11-13
Run Drupal tests as part of `grunt: test`

jhedstrom:
This is a work in progress. See the `@todo`s for next-steps.

### DIFF
--- a/lib/drupal.js
+++ b/lib/drupal.js
@@ -32,7 +32,7 @@ module.exports = function(grunt) {
   };
 
   module.majorVersion = function() {
-    var version = process.env.GDT_DRUPAL_CORE || grunt.config('drupal.drupal-version') || '',
+    var version = grunt.config('drupal.drupal-version') || '',
       versionParts = version ? version.match(/^\d+/) : '';
     return versionParts.length ? parseInt(versionParts[0]) : 0;
   };
--- a/lib/drupal.js
+++ b/lib/drupal.js
@@ -32,7 +32,7 @@ module.exports = function(grunt) {
   };
 
   module.majorVersion = function() {
-    var version = grunt.config('drupal.drupal-version') || '',
+    var version = process.env.GDT_DRUPAL_CORE || grunt.config('drupal.drupal-version') || '',
       versionParts = version ? version.match(/^\d+/) : '';
     return versionParts.length ? parseInt(versionParts[0]) : 0;
   };
--- a/tasks/test.js
+++ b/tasks/test.js
@@ -1,6 +1,7 @@
 module.exports = function(grunt) {
 
   var Drupal = require('../lib/drupal')(grunt);
+  var Help = require('../lib/help')(grunt);
 
   // Default to Drupal 8 unless Drupal 7 is explicitly defined.
   var script = Drupal.majorVersion() == 7 ? './scripts/run-tests.sh' : './core/scripts/run-tests.sh';
--- a/tasks/test.js
+++ b/tasks/test.js
@@ -1,0 +1,39 @@
+module.exports = function(grunt) {
+
+  grunt.config('test', {
+    behat: true,
+    drupal: []
+  });
+
+  grunt.config(['shell', 'testdrupal'], {
+    // @todo Only make this available on Drupal 8+, or change command.
+    // @todo Make configurable.
+    // @todo The phpunit binary (./vendor/bin/phpunit) needs to be executable, GDT changes this somehow during build.
+    // @todo Need to incorporate actual site URL in this command.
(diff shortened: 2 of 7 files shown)


tasks/test.js
```diff
@@ -0,0 +1,39 @@
...
+
+  grunt.config('test', {
+    behat: true,
+    drupal: []
+  });
+
+  grunt.config(['shell', 'testdrupal'], {
+    // @todo Only make this available on Drupal 8+, or change command.
+    // @todo Make configurable.
+    // @todo The phpunit binary (./vendor/bin/phpunit) needs to be executable, GDT changes this somehow during build.
```
arithmetric: @jhedstrom This reminded me of the `mode` option on grunt-contrib-copy. #211 should resolve the removal of the executable mode.

jhedstrom:
Remaining `@todo`s have been removed. Let's see if this goes green.

jhedstrom:
Looks like all the Drupal 7 tests are failing since it isn't properly detecting the version.

lib/drupal.js
```diff
@@ -32,7 +32,7 @@ module.exports = function(grunt) {
   };
 
   module.majorVersion = function() {
-    var version = grunt.config('drupal.drupal-version') || '',
+    var version = process.env.GDT_DRUPAL_CORE || grunt.config('drupal.drupal-version') || '',
```
grayside: When would you need to use the env var?

tasks/test.js
```diff
@@ -0,0 +1,45 @@
...
+    command: grunt.config.get('config.testing.drupal.command') || 'php ' + script + ' --verbose --color --concurrency 4 --directory ' + testLocation + ' --url ' + domain,
+    options: {
+      execOptions: {
+        cwd: '<%= config.buildPaths.html %>'
+      }
+    }
+  });
+
+  // @todo Would rather display as 'alias for test:behat, test:drupal, etc'.
+  grunt.registerMultiTask('test', 'Run all configured tests.', function(args) {
```
grayside: Run with `grunt test:behat`, the args value will be set to `behat`. This might be slightly more clean than the current `self.target`.

tasks/test.js
```diff
@@ -0,0 +1,45 @@
...
+    var self = this;
+    switch (self.target) {
```
grayside: This breaks simply running `grunt test`. Shouldn't it either run behat by default or run all available targets?

tasks/test.js
```diff
@@ -0,0 +1,45 @@
...
+
+  grunt.config('test', {
+    behat: true,
+    drupal: {
+      command: []
+    }
+  });
+
+  grunt.config(['shell', 'testdrupal'], {
+    command: grunt.config.get('config.testing.drupal.command') || 'php ' + script + ' --verbose --color --concurrency 4 --directory ' + testLocation + ' --url ' + domain,
```
grayside: This is primarily to specify the PHP path? Should we make that a globally used "Magic Config" via lib/init.js?

lib/drupal.js
jhedstrom: Our generators don't currently set `drupal-version`, so among other places, GDT's tests use this variable.

grayside: My understanding is the current use of this variable is to route the tests to determine which set of supporting assets to use. For functional purposes we take the version from the build Drupal core codebase. Are we using the env var for something operational? I ask this more to ensure I understand than to object, though it probably bears documentation if we expect this to be used outside of the GDT self-tests.

jhedstrom: I couldn't actually find anywhere else where we are needing the version (drush usually handles that). We could alternatively use drush here to determine the version (as a 3rd fallback)?

grayside: Drush is already used to determine this. drupal.drupal-version is set by dumping `drush status --format=json` into grunt.config('drupal') earlier in this file.

jhedstrom: hmm, I need to revisit why D7 tests were failing without this change then.

tasks/test.js
jhedstrom: There may be other reasons to customize the command (eg, arguments and options). Unless I'm missing something, the `grunt.shell` task this is extending doesn't appear to take options/args as separate config.

arithmetric:
This looks pretty good.

From the Travis tests, it looks like there is an issue on D7 with running SimpleTest I assume:
```
Running "shell:testdrupal" (shell) task
  ERROR: Unknown argument '--directory'.
```

See: https://travis-ci.org/phase2/grunt-drupal-tasks/jobs/111807992#L1924

jhedstrom:
Apparently D7 doesn't support the directory option, so it's either be all or manually specifying custom test groups. I think we should just make this a D8-only task for now (D7 has gotten along thus far without, and projects that want to write d7 simpletests can add it as a custom task).

PR #210, opened 2015-11-20
Ensuring a config setting of "eslint: false" can disable it

EvanLovely:
If `Gruntconfig.json` has `eslint: false` it does not turn it off; this fixes that. The check was looking to see if it wasn't undefined; both `true` and `false` are not undefined... so this moves it to a truthy check.

### DIFF
--- a/lib/drupal.js
+++ b/lib/drupal.js
@@ -68,7 +68,7 @@ module.exports = function(grunt) {
   };
 
   module.drushPath = function() {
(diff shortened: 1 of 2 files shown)


EvanLovely:
How did you comment 14 seconds after I made this?!? Who are you, The Flash? ;)

grayside:
:zap:

EvanLovely:
Added phpcs and phpmd in a similar fashion.

grayside:
https://github.com/phase2/grunt-drupal-tasks/blob/master/lib/drupal.js#L71

PR #211, opened 2015-11-21
Preserve file modes on copy tasks

arithmetric:
Configuring the copy tasks to preserve file mode with 'mode: true'.

By default, grunt-contrib-copy sets the mode to 0644 when it copies files. See: https://github.com/gruntjs/grunt-contrib-copy#mode

### DIFF
--- a/tasks/copy.js
+++ b/tasks/copy.js
@@ -8,37 +8,43 @@ module.exports = function(grunt) {
    */
   grunt.loadNpmTasks('grunt-contrib-copy');
   grunt.config('copy.static', {
+    options: {
+      mode: true
+    },
     files: [
       {
         expand: true,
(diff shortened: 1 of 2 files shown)


jhedstrom:
:+1: this was causing issues when trying to run phpunit.

grayside:
If there's no case where we'd want to whitewash the mode, why not use the [global options property](http://gruntjs.com/configuring-tasks#options) rather than configure per target?

arithmetric:
@grayside Since we have copy-related tasks in a few files, I think we may lose some transparency on how tasks are configured. If you'd like to propose a change along these lines, I'd happily review.

grayside:
Fair argument. I think this is a good change even without extra-DRY config.

PR #214, opened 2015-12-02
Fix null error with eslint exclusion from theme operations

grayside:
Two new enhancements to the eslint integration from v0.9 has been uncovered for otherwise empty projects configured with a theme validate task. This results in breaking the build process when the quality.js file is loaded.

The two changes are:
* Only define the eslint task if there are files that match the configured file mask for eslint processing.
* Exempt theme JS from eslint processing by Grunt Drupal Tasks if the theme has a configured validate script in the project's Gruntconfig.json.

In a newly initialized project, there are no files to process so the eslint task is not defined. However, the theme exemption code assumes the eslint task exists.

### Fix for the Bug

* Retain the definition of theme-specific analyze and validate tasks where they are.
* Add a new code block as part of the eslint task definition to exempt theme files before the file mask is checked.

### Domino Enhancements

As a result of this change, there are two minor improvements.

1. If the theme has an analyze task defined, it will exempt the theme from the Grunt Drupal Tasks analyze.
2. If the only JS files in the codebase are part of an exempt theme, it will now correctly identify that eslint does not need to be configured by Grunt Drupal Tasks.

### Related Bugs Not Fixed

`eslint:validate` and `eslint:analyze` can target different files per the theme exemption, but only the validate tasks files mask is checked to determine if eslint should be configured at all for Grunt Drupal Tasks.

@EvanLovely @arithmetric @tekante

### DIFF
--- a/tasks/quality.js
+++ b/tasks/quality.js
@@ -126,16 +126,31 @@ module.exports = function(grunt) {
     analyze.push('phpmd:custom');
   }
 
+  var themes = grunt.config('config.themes');
   if (grunt.config.get('config.eslint')) {
     var eslintConfig = grunt.config.get('config.eslint'),
       eslintTarget = eslintConfig.dir || [
           '<%= config.srcPaths.drupal %>/**/*.js',
           '!<%= config.srcPaths.drupal %>/sites/**/files/**/*.js'
         ],
+      eslintTargetAnalyze = eslintTarget,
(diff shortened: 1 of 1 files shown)


arithmetric:
I was able to reproduce the eslint:validate issue with a clean install, and updating GDT to use this branch fixed the issue. So LGTM :+1:

PR #223, opened 2015-12-09
Zombie Update [WIP]

grayside:
Setting the wheels in motion to update Zombie from 2.* to 4.*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ addons:
    - gcc-4.8
    - g++-4.8
 env:
-  - NVM_NODE_VERSION="0.12" GDT_DRUPAL_CORE="7" GDT_TEST_URL="http://127.0.0.1:8080/misc/drupal.js" CXX=g++-4.8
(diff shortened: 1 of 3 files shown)


arithmetric:
Zombie 4 requires Node.js 4, so this branch includes changes to require Node.js 4 in the package.json and remove Travis CI tests for Node.js 0.12.

arithmetric:
From the Travis tests, it looks like the D7 Behat JavaScript test are failing, while the D8 Behat JavaScript test is passing.

D7 failure: https://travis-ci.org/phase2/grunt-drupal-tasks/jobs/111881012#L2021

D8 success: https://travis-ci.org/phase2/grunt-drupal-tasks/jobs/111881013#L1911

The D7 Behat JavaScript test relies on the Overlay module, which may not be the best test case.

arithmetric:
Currently, Zombie 4 works with Node 4/npm 2, but fails with Node 5/npm 3 with the error:
```
Error: Cannot find module 'zombie/node_modules/tough-cookie'
```

This can be fixed with the following PR: https://github.com/minkphp/MinkZombieDriver/pull/168

arithmetric:

This test run has an example of the fix with the PR to MinkZombieDriver mentioned above: https://travis-ci.org/phase2/grunt-drupal-tasks/builds/121827772

PR #224, opened 2015-12-14
increasing php quality command speed 

EvanLovely:
Ensures Pattern Lab is not included then excluded in glob. Changes speed of `grunt phplint` from 17s to ~9s.

### DIFF
--- a/tasks/quality.js
+++ b/tasks/quality.js
@@ -26,18 +26,16 @@ module.exports = function(grunt) {
   var analyze = [];
 
   var defaultPatterns = [
-    '<%= config.srcPaths.drupal %>/**/*.php',
-    '<%= config.srcPaths.drupal %>/**/*.module',
-    '<%= config.srcPaths.drupal %>/**/*.inc',
-    '<%= config.srcPaths.drupal %>/**/*.install',
-    '<%= config.srcPaths.drupal %>/**/*.profile',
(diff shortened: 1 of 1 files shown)


grayside:
This sounds like a great improvement. However, why is Pattern Lab code on this scale included under `src/`?

grayside:
Looking at the details, I don't think we should exempt PHP code in themes from the project coding standards.

EvanLovely:
The Pattern Lab core code being committed is a long story and I'd be happy to tell you over a video chat, but it's something we don't like either and have a plan for removing. 

However, you'll see that PHP code is being included, I've added all files in `theme_name/templates/` or in `theme_name/src/` (a potential future location) in the theme and `theme_name/template.php`. Those source encompass all places that PHP code exists in the theme.

grayside:
Ah! That was my misread. I guess my remaining concern is the specificity of PHP in the theme seems like it might be very tied to how the PLS* projects are structuring things. Is that enough of a ubiquitous practice that it should be proscribed as a GDT default vs. placed in Gruntconfig via the generator?

EvanLovely:
I feel it is general enough and give the fact that the array of locations can be overridden in Gruntconfig.json, we should be good to go. This PR is just for the default settings.

arithmetric:
Thanks for this PR. This was built upon in #233 which is now merged, so closing this.

PR #226, opened 2016-01-21
Store this.name in a variable

mirie:
The scope of "this" changes in the function defined in Drupal.loadDatabaseConnection(), so this.name is undefined and the pre- and post-install tasks are never added to the task list.


jhedstrom:
:+1: Thanks!

PR #227, opened 2016-01-25
Fixing bug that prevents phpcs and eslint from running.

arithmetric:
The fix is to ensure that template variables are processed before the source paths are expanded.


jhedstrom:
:+1: I tested this and the `validate` command works as expected.

PR #230, opened 2016-01-28
Defer task identification on validate and analyze operations.

grayside:
Reorganize quality.js load to configure the various tasks, but not actually identify which will run until the analyze or validate task callback is invoked by grunt. This way operations like the file check to see if eslint should run won't be operated until you run `grunt validate`, rather than running freely on such commands as `grunt help`.

### DIFF
--- a/tasks/quality.js
+++ b/tasks/quality.js
@@ -188,7 +188,7 @@ module.exports = function(grunt) {
     }
     var eslint = grunt.config.get('eslint.validate');
     if (eslint) {
-      if (filesToProcess(eslint.validate)) {
+      if (filesToProcess(eslint)) {
         validate.push('eslint:validate');
       }
     }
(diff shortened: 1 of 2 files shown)


grayside:
Not sure why this failed testing after I "fixed" eslint. This error is similar to something I've seen in Jenkins: https://travis-ci.org/phase2/grunt-drupal-tasks/jobs/105563850#L1512

tasks/quality.js
```diff
@@ -156,30 +145,19 @@ module.exports = function(grunt) {
...
-        validate: eslintTarget,
-        analyze: {
-          options: {
-            format: 'checkstyle',
-            outputFile: '<%= config.buildPaths.reports %>/eslint.xml'
-          },
-          src: eslintTargetAnalyze
-        }
-      });
-      validate.push(eslintName + ':validate');
```
arithmetric: @grayside It looks like eslint issues are causing the build to abort because the use of `eslintName` was lost, which includes the "force:" prefix when `ignoreError` is set. I'll try to re-introduce this use.

grayside: Nice catch, and it looks like success!

PR #231, opened 2016-01-28
Pattern Reuse for eslint and phplint

grayside:
This furthers the work in #230 by allowing our initial filescan checking to see if there are files to be linted to reuse the identified files in the linting process itself. This seems like it might be a micro-optimization, as I suspect it's passing the list of files through the globbing system, which reduces the potential gain of what could be a simple passthru. That said, if it works every few seconds counts.

### DIFF
--- a/example/.eslintrc
+++ b/example/.eslintrc
@@ -23,7 +23,6 @@
     "no-use-before-define": 0,
     "consistent-return": 0,
     "no-constant-condition": 0,
-    "no-comma-dangle" : 2,
     "no-catch-shadow" : 2
   }
 }
--- a/tasks/quality.js
+++ b/tasks/quality.js
@@ -175,20 +175,24 @@ module.exports = function(grunt) {
       return grunt.template.process(item);
     });
 
-    // If this is evaluated to truthy at least one file matched.
-    return grunt.file.expand(paths).length;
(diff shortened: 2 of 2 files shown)


arithmetric:
To restate, the outcome of this PR is that the source files configured to be validated/analyzed with file pattern globs is replaced by the normalized, explicit list of files identified by filesToProcess(). This results in eliminating file scan process to resolve patterns that would be performed by the Grunt plugins for validate/analyze tasks.

grayside:
> To restate, the outcome of this PR is that the source files configured to be validated/analyzed with file pattern globs is replaced by the normalized, explicit list of files identified by filesToProcess(). This results in eliminating file scan process to resolve patterns that would be performed by the Grunt plugins for validate/analyze tasks.

Yes. Thus each task will only filescan once. However, I found it to be of limited timing utility because I suspect it's still putting each file through the globbing library, so now instead of ~8 slow patterns, we have ~50+ fast ones.

arithmetric:
It looks like #233 introduces this approach as well. Will that PR implement the main outcomes of this one?

grayside:
I expect @mdeltito anticipated this going forward and wanted to head off merge conflicts. We can close this in favor of the other.

arithmetric:
Closing in favor of #233.

PR #233, opened 2016-02-05
Scale back default validation patterns for phpcs/phplint and eslint

mdeltito:
This adds a more restrictive set of default validation patterns, avoiding excessive globbing particularly in locations that might contain node_modules or similarly large directory trees. Although the existing patterns exclude these large directories from linting, patterns that match these directories (inclusive or exclusive) incur a filesystem stat which can decrease performance, especially with the recursive nature of node_modules with npm < 3. 

The summary of changes is:

1. No top-level globstar in any pattern to avoid recursing through `src/themes`, exclusive or inclusive
2. Scope php validation to the modules & profiles directories **only**.
3. Limit eslint to `themes/<theme_dir>/js` to avoid globbing through node_modules and bower_components.
4. Adjusted some defaults for common false positives and generated code.

These changes err on the side of scanning as few directories as possible while still picking up the files that we actually care about. Project-specific tweaks can add more targeted patterns as needed.

### DIFF
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -373,10 +373,10 @@ This is an example of the settings for the validate tasks:
   }
   "phplint": {
     "dir": [
-      '<%= config.srcPaths.drupal %>/themes/atmos_energy/template.php',
-      '<%= config.srcPaths.drupal %>/themes/atmos_energy/templates/**/*.php',
-      '<%= config.srcPaths.drupal %>/themes/atmos_energy/includes/**/*.{inc,php}',
       '<%= config.srcPaths.drupal %>/sites/**/*.{php,inc}',
+      '<%= config.srcPaths.drupal %>/themes/*/template.php',
+      '<%= config.srcPaths.drupal %>/themes/*/templates/**/*.php',
+      '<%= config.srcPaths.drupal %>/themes/*/includes/**/*.{inc,php}',
       '<%= config.srcPaths.drupal %>/{modules,profiles,libraries,static}/**/*.{php,module,inc,install,profile}',
       '!<%= config.srcPaths.drupal %>/{modules,profiles,libraries,static}/**/*.{box,pages_default,views_default,panelizer,strongarm}.inc',
       '!<%= config.srcPaths.drupal %>/{modules,profiles,libraries,static}/**/*.features.*inc',
@@ -403,7 +403,7 @@ review by eslint. The following is used by default:
   "eslint": {
     "dir": [
       'src/themes/*/js/**/*.js',
-      'src/{modules,profiles}/**/*.js'
+      'src/{modules,profiles,libraries}/**/*.js'
     ]
   }
 }
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -363,9 +363,8 @@ This is an example of the settings for the validate tasks:
 {
   "eslint": {
     "dir": [
-      '<%= config.srcPaths.drupal %>/**/*.js',
-      '!<%= config.srcPaths.drupal %>/**/bower/**/*.js',
-      '!<%= config.srcPaths.drupal %>/sites/**/files/**/*.js'
+      '<%= config.srcPaths.drupal %>/themes/*/js/**/*.js',
+      '<%= config.srcPaths.drupal %>/{modules,profiles}/**/*.js'
     ]
   },
   "phpcs": {
@@ -374,12 +373,11 @@ This is an example of the settings for the validate tasks:
   }
   "phplint": {
     "dir": [
-      "<%= config.srcPaths.drupal %>/**/*.php",
-      "<%= config.srcPaths.drupal %>/**/*.module",
-      "<%= config.srcPaths.drupal %>/**/*.inc",
-      "<%= config.srcPaths.drupal %>/**/*.install",
-      "<%= config.srcPaths.drupal %>/**/*.profile",
-      "!<%= config.srcPaths.drupal %>/**/vendor/**",
+      '<%= config.srcPaths.drupal %>/{modules,profiles}/**/*.{php,module,inc,install,profile}',
+      '!<%= config.srcPaths.drupal %>/{modules,profiles}/**/*.{box,pages_default,views_default,panelizer,strongarm}.inc',
+      '!<%= config.srcPaths.drupal %>/{modules,profiles}/**/*.features.*inc',
+      '!<%= config.srcPaths.drupal %>/{modules,profiles}/**/*.tpl.php',
+      '!<%= config.srcPaths.drupal %>/{modules,profiles}/**/vendor/**'
     ]
   }
 }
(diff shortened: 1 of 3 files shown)


tasks/quality.js
```diff
@@ -46,9 +40,7 @@ module.exports = function(grunt) {
   validate.push('phplint:all');
 
   if (grunt.config.get('config.phpcs')) {
-    var phpcs = grunt.config.get('config.phpcs.dir') || [
-        '<%= config.srcPaths.drupal %>/**/*.css'
```
mdeltito: Killed this off to prevent globbing. It seems generally undesirable in most cases to run phpcs on css files anyway, since they are almost always generated.

CONFIG.md
```diff
@@ -363,9 +363,8 @@ This is an example of the settings for the validate tasks:
 {
   "eslint": {
     "dir": [
-      '<%= config.srcPaths.drupal %>/**/*.js',
-      '!<%= config.srcPaths.drupal %>/**/bower/**/*.js',
-      '!<%= config.srcPaths.drupal %>/sites/**/files/**/*.js'
+      '<%= config.srcPaths.drupal %>/themes/*/js/**/*.js',
+      '<%= config.srcPaths.drupal %>/{modules,profiles}/**/*.js'
```
grayside: This should also be retained in libraries and static.

CONFIG.md
```diff
@@ -374,12 +373,11 @@ This is an example of the settings for the validate tasks:
   }
   "phplint": {
     "dir": [
-      "<%= config.srcPaths.drupal %>/**/*.php",
-      "<%= config.srcPaths.drupal %>/**/*.module",
-      "<%= config.srcPaths.drupal %>/**/*.inc",
-      "<%= config.srcPaths.drupal %>/**/*.install",
-      "<%= config.srcPaths.drupal %>/**/*.profile",
-      "!<%= config.srcPaths.drupal %>/**/vendor/**",
+      '<%= config.srcPaths.drupal %>/{modules,profiles}/**/*.{php,module,inc,install,profile}',
```
grayside: Should be applicable to sites, static, libraries without impeding theme subdirectories, right?

CONFIG.md
```diff
@@ -374,12 +373,11 @@ This is an example of the settings for the validate tasks:
   }
   "phplint": {
     "dir": [
-      "<%= config.srcPaths.drupal %>/**/*.php",
```
grayside: I was initially unsure of removing themes, but in practice I have found a default assumption that rigid coding standards can be applied to templates is fairly difficult. In that light, better to have a project opt-in.

grayside: (Above comment to PHPCS).

PHP-lint, on the other hand, can prevent fatal errors, and would be nice to retain for themes. Can we not assume PHP in a template.php and templates directory?

CONFIG.md
mdeltito: I purposefully removed libraries and static from the defaults and from the docs, as it seems rare that you would want to be scanning third-party libraries (and rarer that you are writing your own in this location). I figured it would be reasonable to have these directories be opt-in, but I can add them back if that's a bad assumption.

CONFIG.md
mdeltito: Same as above for elint. If we want to include any theme scanning, it would have to be explicit subdirectories and/or files (template.php) to avoid glob overhead.

CONFIG.md
mdeltito: I think we can at minimum include linting for template.php, and perhaps the best-practice naming convention for a templates and includes directory. Open to suggestions, but it would be something like:

`'<%= config.srcPaths.drupal %>/themes/' + key + '/template.php',`
`'<%= config.srcPaths.drupal %>/themes/' + key + '/templates/**/*.php',`
`'<%= config.srcPaths.drupal %>/themes/' + key + '/includes/**/*.{inc,php}'`

CONFIG.md
grayside: libraries is for custom libraries, be it PHP or JS, and static is for new files to be copied into document root, which could include custom bootstrap scripts.

CONFIG.md
grayside: That seems fair for phplint to help keep things stable.

mdeltito:
Updated to address feedback. I did some similar refactoring as in #231 to prevent the need for multiple calls to `grunt.file.expand` between config and runtime. It looks like this may need to be rebased on #230 and #231 once that is in anyway once those are in.

CONFIG.md
grayside: Still not scanning libraries from eslint.

grayside: Ah, this is the CONFIG.md, missed that.

CONFIG.md
```diff
@@ -401,8 +402,8 @@ review by eslint. The following is used by default:
 {
   "eslint": {
     "dir": [
-      '<%= config.srcPaths.drupal %>/**/*.js',
-      '!<%= config.srcPaths.drupal %>/sites/**/files/**/*.js'
+      'src/themes/*/js/**/*.js',
+      'src/{modules,profiles}/**/*.js'
```
grayside: libraries here

tasks/quality.js
```diff
@@ -131,12 +139,11 @@ module.exports = function(grunt) {
...
   }
 
-  var themes = grunt.config('config.themes');
   if (grunt.config.get('config.eslint')) {
     var eslintConfig = grunt.config.get('config.eslint'),
       eslintTarget = eslintConfig.dir || [
-          '<%= config.srcPaths.drupal %>/**/*.js',
-          '!<%= config.srcPaths.drupal %>/sites/**/files/**/*.js'
+          '<%= config.srcPaths.drupal %>/themes/*/js/**/*.js',
+          '<%= config.srcPaths.drupal %>/{modules,profiles}/**/*.js'
```
grayside: libraries here.

tasks/quality.js
```diff
@@ -25,31 +25,38 @@ module.exports = function(grunt) {
...
+    '!<%= config.srcPaths.drupal %>/{modules,profiles,libraries,static}/**/*.{box,pages_default,views_default,panelizer,strongarm}.inc',
+    '!<%= config.srcPaths.drupal %>/{modules,profiles,libraries,static}/**/*.features.*inc',
+    '!<%= config.srcPaths.drupal %>/{modules,profiles,libraries,static}/**/vendor/**'
   ];
 
+  // Include common sites and theme locations in phplint validation.
+  var phplintPatterns = defaultPatterns.slice(0);
+  phplintPatterns.unshift('<%= config.srcPaths.drupal %>/sites/**/*.{php,inc}');
+
+  for (var key in themes) {
```
grayside: If a theme is not wired into the GDT config, we still want to scan it. The theme configuration is intended to target themes taking over more of their own functionality.

Maybe this works as:
```
     '<%= config.srcPaths.drupal %>/themes/*/template.php',
     '<%= config.srcPaths.drupal %>/themes/*/templates/**/*.php',
     '<%= config.srcPaths.drupal %>/themes/*/includes/**/*.{inc,php}'
```

mdeltito:
One more time, with libraries. Also unscoped theme phplint from explicitly configured themes.

grayside:
This is looking ready for conflict resolution.

arithmetric:
I'm ready to merge this once the conflicts are resolved.

mdeltito:
I ended up re-implementing the new patterns based on @grayside's work in #230, and thus was able to merge and test #231 with the new defaults here. Also fixed the remaining minor issues noted in #231.

grayside:
:+1: since it's heavily rebased onto my other PRs I think @arithmetric should hit the button.

PR #235, opened 2016-02-12
Use rsync in place of copy to move Drush Make output to build.html.

grayside:
This change swaps copy:tempbuild for rsync:tempbuild. In local environments, across NFS mounts, and so on a large Drupal site might take 10 minutes for a copy operation. This accelerates that timing to 2-3 minutes.

The grunt-rsync wrapper uses https://github.com/jedrichards/rsyncwrapper under the hood, which mentions rsync (>=2.6.9) is a dependency. I'm not sure of the implications for windows compatibility.

This is a first step towards switching the grunt package command to use rsync. The existing implementation is not able to package a fully assembled atrium site without modifying the max stacksize limits of the grunt process, and also benefits from this speedup.


grayside:
Failure relates to shrinkwrap being out of date, when I run `npm shrinkwrap` locally it switches out all the dependencies along these lines:

```
-      "from": "argparse@>=1.0.2 <2.0.0",
+      "from": "https://registry.npmjs.org/argparse/-/argparse-1.0.3.tgz",
```

Using npm v2.11.3.

arithmetric:
@grayside I ran `npm update; npm shrinkwrap` to update the shrinkwrap json file.

grayside:
Hm, looks like you got the same result. Maybe npm is shrinkwrapping differently.

grayside:
Tests are passing, which leaves the question of rsync dependency.

arithmetric:
It looks like rsync 2.6.9 pretty old (released in 2006), so that requirement may not be too challenging. However, I'm not sure about the general availability of rsync on Windows.

What do you think about making it configurable (maybe enabled by default, but able to be reverted back to the copy operation by adding rsync:false to the Gruntconfig.json)?

grayside:
Awkward git mess, creating a new PR.

PR #238, opened 2016-02-23
Updating tests and example to use Grunt 1.0.0-rc1

arithmetric:
Opening for follow up work from #236.


arithmetric:
Closing in favor of #243 and #244.

PR #240, opened 2016-02-25
Updating dependencies, related fixes

arithmetric:


arithmetric:
Note: As of v2.0.0, grunt-force-task supports only Node.js v4.0.0 and later, so updating this dependency will have to wait until #241.

PR #243, opened 2016-02-29
Removing peer dependencies

arithmetric:
Since the role of peer dependencies is changing in npm and they interfere with new releases of Grunt (prompted by #236), let's remove the peer dependencies section of package.json.

The example package.json includes a dependency on grunt, so the need for the peer dependency in Grunt Drupal Tasks is not clear.


grayside:
Almost seems like grunt should be an engine instead, but let's get this in to stop the warning spam and sort that out in follow-up.

PR #245, opened 2016-03-01
Faster rsync for tempbuild step in non-Windows environments.

grayside:
Picking up from https://github.com/phase2/grunt-drupal-tasks/pull/235.

This change preserves copy:tempbuild, uses it on Windows, and on non-Windows environments switches to rsync. I thought about fancier rsync detection, but came to the conclusion I couldn't effectively test the detection mechanism itself.

As part of making this logic, I moved the configuration of drushmake build steps into the global config for our "drushmake" task. This means any use of `grunt.config.set('drushmake.options.tasks', {})` can overwrite the build process, a bit more flexible than the current requirement in bouncing around the registered drushmake task.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -7,5 +7,10 @@ module.exports = {
     return paths.map(function(value, index, arr) {
       return baseUrl + value;
     });
+  },
+  canRsync: function() {
+    // Enable process.platform to be mocked in options for testing.
+    var platform = options.platform || process.platform;
+    return platform === "win32";
   }
 };
--- a/lib/util.js
+++ b/lib/util.js
@@ -11,6 +11,6 @@ module.exports = {
   canRsync: function() {
     // Enable process.platform to be mocked in options for testing.
     var platform = options.platform || process.platform;
-    return platform === "win32";
+    return platform !== "win32";
   }
 };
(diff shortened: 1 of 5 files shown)


arithmetric:
Thanks for this update.

I see this error in a few of the Travis tests:

```
Running "drushmake:default" (drushmake) task
Warning: Task "rsync.tempbuild" not found. Use --force to continue.
Aborted due to warnings.
```

I think the following bits got missed in this branch:

- Adding grunt-rsync in package.json
- Loading with grunt.loadNpmTasks

lib/util.js
```diff
@@ -7,5 +7,10 @@ module.exports = {
     return paths.map(function(value, index, arr) {
       return baseUrl + value;
     });
+  },
+  canRsync: function() {
+    // Enable process.platform to be mocked in options for testing.
+    var platform = options.platform || process.platform;
+    return platform === "win32";
```
arithmetric: Shouldn't this be `platform !== "win32"`?

arithmetric:
After the updates, the tests are still failing with this error:

```
Running "drushmake:default" (drushmake) task
Warning: Task "rsync.tempbuild" not found. Use --force to continue.
```

One thing that caught my eye was including options in the config key here rather than as part of the object being set: `grunt.config('rsync.tempbuild.options', {`

arithmetric:
There's a new error: https://travis-ci.org/phase2/grunt-drupal-tasks/jobs/113320963#L1822

grayside:
Failing on Node 5 only.

PR #248, opened 2016-03-01
Adding eslint for GDT

arithmetric:
Adding eslint configured to use the Google JavaScript standards. Excludes node_modules and test assets and output.

This is a fairly opinionated standard and will require a fair amount of initial code clean up. I think it will help us achieve a more consistent code style and potentially catch errors that are difficult to spot.


arithmetric:
Closing in favor of #283.

PR #249, opened 2016-03-02
Use PhantomJS instead of Zombie

jhedstrom:
Initial research and work for issue #247.


jhedstrom:
Tests are failing due to https://github.com/ariya/phantomjs/issues/11637. Other drivers it seem will happily click on the first link with a matching title (eg, the 'my account' link is on the page twice, the first one hidden via js, the second always available).

I'm going to update the test to work around this, but that limitation may be enough reason not to default to phantomjs.

PR #254, opened 2016-03-05
Release Packaging Support for Acquia

grayside:
Took #195 and tried to take it home. Looks like local checkout of a PR you don't own the branch on leads to a break from the original PR.

The core change is to allow specifying different package destinations via Gruntconfig or a CLI flag. The next step in evolving package is to flip back from the "name" convention I swapped it to here and go with a formal multi-target task so different packaging options can more easily be supported.

Changes in this version also include an entry in grunt help for the new option and some basic test coverage, which I'm waiting on Travis to confirm.


grayside:
I expected that run to leave us to Node 5 failures and all else green. Does create_working_copy.sh not install devDependencies?

arithmetric:
@grayside It seems that `npm install <package>` does not install dev dependencies; rather they're only installed when you run `npm install` plain on the package.json in the current directory.

To install dev dependencies for the tests, one option is to run `npm install` in the GDT dependency directory after it is installed in the working copy. For example: https://gist.github.com/arithmetric/ff21b2377790f205e06c

(I tested the `--dev` and `--only=dev` options: the former installs dependencies recursively, which is way too expansive and is being deprecated, and the latter option is only available in npm 3.)

arithmetric:
This has been merged into the v1.0.0-pre branch.

PR #258, opened 2016-03-11
Use --root to set docroot path in lieu of working directory.

grayside:
Lately I've been finding that drush cannot always recognize Drupal based on the current working directory. This is local to my machine and to several other developers I've been working with. Very strange behavior, and this change does not fix it. However, I have noted that if you use the --root option you can sidestep this problem, which should at least mitigate the problem in our primary tooling while the real issue is addressed.

### DIFF
--- a/tasks/install.js
+++ b/tasks/install.js
@@ -14,37 +14,33 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-drush');
 
   var Help = require('../lib/help')(grunt),
-    Drupal = require('../lib/drupal')(grunt);
-
-  var path = require('path'),
+    Drupal = require('../lib/drupal')(grunt),
+    path = require('path'),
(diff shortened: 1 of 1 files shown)


grayside:
Drush is complaining about PHP 5.4

> `Your command line PHP installation is too old. Drush requires at 
least PHP 5.5.0. To suppress this check, set the environment variable
DRUSH_NO_MIN_PHP=1`

arithmetric:
@grayside To confirm, above you said that this change does not fix the issue you have. Is it correct that this change hasn't fixed the issue on any environments?

Could you update this with master to see if the test errors are fixed since #259 is now merged?

grayside:
@arithmetric updated the codebase.

This fixes the issue of grunt install failing by working around the bug I observed with drush not able to identify a drupal docroot from the current working directory.

PR #260, opened 2016-03-11
Drop support for PHP 5.4 due to emerging Drush incompatibility.

grayside:
Per a Drush error on recent test runs, PHP 5.4 support is being dropped from Drush. This PR pulls it from travis testing by way of dropping it in GDT as well. If we do not feel PHP 5.4 support can be let go, we should instead identify what version of Drush will be safe to use and plan to pin that in composer.json.

See also #259 

@jhedstrom

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: php
 php:
-  - 5.4
   - 5.5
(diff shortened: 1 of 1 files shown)


arithmetric:

PR #261, opened 2016-03-14
Remove theme validate from validate:staged.

grayside:
`grunt validate:staged`, which is used by the githooks system, chokes if a theme is configured because there's nothing that can be done with `staged:themes:<name>:validate`.

This change causes tasks that are prefixed `themes:` to be pulled from the `validate:staged` handling.


grayside:
The failure of the push itself is do to a timeout in one of the builds. Let's see what the PR test reports.

grayside:
Test passes.

PR #267, opened 2016-03-28
Improve performance and cacheability of composer install.

### LINKED ISSUE #253: Use --prefer-dist on composer install
When running composer install, many packages are installed first via an attempt at git checkout. However, on Github and other services, anonymous checkouts are rate limited, and it is easy for a CI server to hit that limit.

Using a distribution package (which many projects have) as the primary candidate let's us avoid hitting up against this rate limit, and avoids round trips just to retrieve access denied when the rate limit has been hit, before composer falls back to a dist approach anyway.

In [reading the documentation](https://getcomposer.org/doc/03-cli.md#install) for how this works, it's not as clean cut as I originally imagined from looking at composer output:

> --prefer-source: There are two ways of downloading a package: source and dist. For stable versions Composer will use the dist by default. The source is a version control repository. If --prefer-source is enabled, Composer will install from source if there is one. This is useful if you want to make a bugfix to a project and get a local git clone of the dependency directly.
> --prefer-dist: Reverse of --prefer-source, Composer will install from dist if possible. This can speed up installs substantially on build servers and other use cases where you typically do not run updates of the vendors. It is also a way to circumvent problems with git if you do not have a proper setup.

So this likely needs some experimentation first.

grayside:
Optimizes performance for rare dependency changes and CI that cleans the vendor directory. Note the --prefer-dist option means attempting to make more use of release tarballs, in the past this could lead to hitting Github API limits, but in the last couple weeks Github decided to remove those limits for use cases such as composer.

Fixes #253 

@jhedstrom @arithmetric

### DIFF
--- a/tasks/composer.js
+++ b/tasks/composer.js
@@ -11,7 +11,15 @@ module.exports = function(grunt) {
     grunt.loadNpmTasks('grunt-composer');
     var Help = require('../lib/help')(grunt);
 
-    grunt.config(['composer', 'install'], {});
+    grunt.config(['composer', 'install'], {
+      options: {
+        flags: [
+          'no-interaction',
+          'no-progress',
+          'prefer-dist'
+        ],
+      }
+    });
 
     Help.add({
       task: 'composer',


jhedstrom:
Makes sense to me. :+1:

grayside:
The test failure appears to be a fluke, re-running again in case it is a repeating fluke.

PR #269, opened 2016-03-30
util.canRsync() is a function

grayside:
As noted by @mdeltito in https://github.com/phase2/generator-gadget/pull/59


arithmetric:
It looks like canRsync had not implemented the options argument as a mock.

Issue #251, opened 2016-03-02
Drop built-in support for Compass


arithmetric:
Now that theme scripts, Grunt, or Gulp tasks can be run as part of GDT theme tasks, the built-in support for Compass can be dropped.

This will allow simplifying the theme task code, and remove the complexity of Compass and dependencies (Ruby, Bundler, etc.) from the test assets and documentation.

PR #270, opened 2016-04-07
Drop built-in support for Compass

arithmetric:
As explained in #251, GDT's built-in support is less useful with support for delegating to theme scripts.


arithmetric:

PR #273, opened 2016-04-12
Improve newer for drushmake performance

grayside:
This builds on #272 .

The newer scanner is current recursively checking the entire src directory (if I read this right) looking for makefiles. This likely explains why projects sometimes have a longish pause toward what seems like the end of the validate step.

I have not come across a use case for needing that recursion, and it's pretty easy to customize the makefiles to scan via `grunt.config.set('drushmake.default.src', [ 'alternate paths'])`.

B/C break, so targeting 1.0.

### DIFF
--- a/tasks/make.js
+++ b/tasks/make.js
@@ -50,7 +50,7 @@ module.exports = function(grunt) {
     default: {
       src: [
         '<%= config.srcPaths.make %>',
-        '<%= config.srcPaths.drupal %>/**/*.{make,make.yml}'
+        '<%= config.srcPaths.drupal %>/*.{make,make.yml}'
       ],
       dest: '<%= config.buildPaths.html %>',
     },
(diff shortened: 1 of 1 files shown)


grayside:
Should have targeted the 1.0.0-pre branch.

tasks/make.js
```diff
@@ -48,7 +48,10 @@ module.exports = function(grunt) {
   // source directory to catch make files included from the primary one.
```
arithmetric: @grayside This comment suggests that the expansive glob pattern is meant to ensure that changes to included make files, especially with the "newer" operation. Will the change in this PR conflict with this purpose?

grayside: Yes, it will no longer descend into directories by default. I'm going to take a smarter approach that will be a better match for something we might do with Composer.

PR #275, opened 2016-04-16
Fix packaging with Grunt v1 and G-D-T v0.11

### LINKED ISSUE #274: `grunt package` is broken with g-d-t 0.11.x and grunt 1.x
When a build is performed with them, sites/all/modules/custom and sites/all/themes/custom are empty.

arithmetric:
Adds the `follow: true` option for node-glob to ensure it follows symlinks into directories for relevant copy tasks. This fixes an issue where the output of `grunt package` did not include files targeted by symlinks, like src/modules from sites/all/modules/custom.

Also adds test coverage for the build and package tasks.

Fixes #274.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,5 +57,6 @@ script:
   - grunt test
   - sleep 1; while (ps aux | grep '[b]ehat' > /dev/null); do sleep 1; done
   - for pid in `ps aux | grep drush | grep runserver | awk '{print $2}'`; do echo "Stopping drush pid $pid"; kill -SIGINT $pid; done;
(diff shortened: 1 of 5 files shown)


grayside:
#254 also has a bunch of test coverage. Other than deciding whether that should go in first, this looks good to me.

grayside:
Other issue looks jammed up, let's get this through, but I'd like to see if there are any test cases we could grab into this PR.

arithmetric:
@grayside Thanks for the review. I'll move forward on a release with just this change, but I support getting #254 into the next release.

PR #277, opened 2016-04-23
Initial take on adding Drupal core composer install for >= 8.1.0

grayside:
This PR checks to see if the version of Drupal is >= 8.1, and if so it adds an extra composer:install operation. The theory of the require statements is to make sure it does not operate unless drushmake and scaffold already ran. This ensures the codebase with it's composer.json is already present, and that the custom modules and libraries from the codebase have been symlinked in for any composer autoloader magic.

I suppose an alternative to adding the semver check would be looking for composer.json and !vendor.

The build process is currently incomplete, now that Drupal 8.1 does not ship the vendor directory pre-installed.

@arithmetric @jhedstrom 

Warning: Untested changeset.

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -54,6 +54,7 @@ module.exports = function(grunt) {
 
   // Wire up the generated docroot to our custom code.
   tasksDefault.push('scaffold');
+  tasksDefault.push('composer-drupal');
 
   if (grunt.config.get(['composer', 'install'])) {
     tasksDefault.unshift('composer:install');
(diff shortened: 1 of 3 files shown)


grayside:
This issue is is slightly mitigated because packaged versions of Drupal core on Drupal.org already include the vendor directory. However, other forks of Drupal Core likely do not use this, nor do git checkouts.

The next hurdle to this issue is that our Drupal version detection is dependent on `drush status`, and that command is dependent on bootstrapping Drupal which requires the composer autoloader. Since problems relying on drush in this way have come up before, maybe we should adjust lib/drupal.js?

arithmetric:
@grayside Now that #284 has been merged, I believe the goals of this PR have been accomplished. Do you see any follow ups needed or can this be closed?

Issue #255, opened 2016-03-05
Revise grunt package to be multi-target, cleaner configuration schema


grayside:
grunt package is versatile, but is fundamentally limited to one type of packaging per project. Further, I think the naming conventions in it's structure are confusing.

1. Revise package to be a multi-target grunt task.
2. Move `:compress` target to `--compress` as a flag.
3. Convert the config schema to something like what's proposed below.
4. Consider extracting the core package functionality to a separate project.

### Proposed configuration schema

The schema below demonstrates something somewhat complete.

```
{
  package: {
    options: {
       src: {
         docroot: {
           include: [ "build/html" ],
           exclude: [ "build/html/xmlrpc.php", "build/html/modules/php" ],
         },
         root: {
           include: [ "*.md" ],
           exclude: [],
         }
       },
       dest: {
         docroot: "",
         root: ""
       },
       compressed: false
    },
    default: {
       "name": "override-target-name"
    },
    archive: {
       compressed: true
    },
    acquia: {
      src: {
        root: {
          include: [ "hooks", "bin" ],
        }
      }
      dest: {
        docroot: "docroot"
      }
    },
    docker: {
      dest: {
        docroot: "/root/var/www",
        root: "/root/var/docs",
      }
    }
  }
}
```

The default configuration GDT would need looks more like this:
```
{
  package: {
    options: {
       src: {
         docroot: {
           include: [ "<%= config.buildPaths.html %>" ],
           exclude: [ "<%= config.buildPaths.html %>/xmlrpc.php", "<%= config.buildPaths.html %>/modules/php" ],
         },
         root: {
           include: [ "*.md" ],
         }
       },
       dest: {
         docroot: "",
         root: ""
       }
    },
    default: {
      name: "package"
    },
  }
}
```
Note that file masks should conform with typical Bash passthru rather than general Grunt file matching, per #252. The specific scheme below will **not** be compatible with grunt systems or plugins that think they know what `src` and `dest` might contain.

The "namespaces under the `options.src` have no special meaning, and merely provide "subtargets" for a given package job. Configuration paths are merged from default into the targets in the case of arrays, and overridden by targets in the case of scalar values.

package can be run as `grunt package:default` for a specific target, or `grunt package:default:root` for a specific namespace. Note that combined with the compress flag this either discards other namespaces or we would need to update files within the archive which sounds potentially messy for cross-compatibility. Each namespace would be a separate rsync or copy operation.

PR #278, opened 2016-04-27
Split up docs and prepare for static site baking

grayside:
This is a pretty significant refactoring and partial rewrite of the documentation to flow along more topical lines. It is a stepping stone towards generating a GDT site along the lines of http://phase2.github.io/devtools.

I'd like to minimize perfecting the docs in this PR, but obviously if there are any major regressions in the quality that should be a blocker.

Early PR to invite feedback while I do some continued polishing. By the conflict, looks like I may have missed a change in CONFIG.md.


arithmetric:
This looks great, I'm ready to merge when the conflict is resolved.

It looks like the differences came from the packaging branch: https://gist.github.com/arithmetric/a4a08d92c0de2da6e0ccb3a88eb2d9ff

grayside:
Conflict resolved. Please note that I am still hoping to introduce some B/C breaks via #255 and rsync-based packaging. Might even have recruited someone to take point on that.

PR #282, opened 2016-05-15
Do not traverse sites sub-directories on quality tasks.

grayside:
We are currently traversing into the site files directory for phplint. Let's stop doing that, as it can be a significant performance overhead for large sites, and in the case of one broken moment in Drupal 8 development, the basis to fail phplint validation on the codebase because a broken PHP file was dropped in the files directory.

Is there a one line way to specify "two levels deep, no more"?

@mdeltito

### DIFF
--- a/tasks/quality.js
+++ b/tasks/quality.js
@@ -36,7 +36,8 @@ module.exports = function(grunt) {
   // Include common sites and theme locations in phplint validation.
   var phplintPatterns = defaultPatterns.slice(0);
   phplintPatterns.unshift.apply(phplintPatterns, [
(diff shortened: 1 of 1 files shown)


arithmetric:
Is excluding the sites/files directory a possibility instead of changing whether other subdirectories are scanned?

grayside:
A recursive scan would spend time loading all files, then the negate would remove them. Or so was my understanding from how the optimization around node_modules played out.

mdeltito:
I found previously that exclusions have the same impact on performance due to how the underlying glob/matching library works. Something like the following incurs a scan of all files in `sites/files`, and then gets expanded to exclude the individual files that match the exclusion pattern:

```
sites
!sites/files
```

Issue #169, opened 2015-06-04
Standardize Gruntconfig.json keys


grayside:
Our Gruntconfig schema has grown organically over the last year, and there are a few inconsistencies that would help folks trying to configure for their projects.
1. Top-level keys should be either globally used configuration or the top-level operations as documented in `grunt help`.
2. Paths to script binaries should be consolidated, for example: `bin: { phpcs: "vendor/bin/phpcs", phpmd: "vendor/bin/phpmd" }`.
3. Grunt plugins that can't use some new generic configuration (e.g., `validate.ignoreExitCode`) should be properties inside the top level operation they are most commonly used in.
4. The analyze task is a bit of a special case as a more elaborate, report-generating version of validate. It should merge it's plugin-specific configuration over the plugin-specific validate configuration. So, if you configured `validate.phpcs.scan` this would be used if `analyze.phpcs.scan` is not set.
5. Properties that are meant to achieve the same goal, such as determining which files to include in a process, should use the same key. I proposed "scan" in (4), a more typical name in Grunt would be simply "files".
- **Pros**:
  - The configuration more closely aligns with our help system and primary concepts of GDT.
  - Greater consistency improves intuitiveness.
- **Cons**:
  - If we have some common configuration to a plugin that we want to make available via different top-level GDT tasks it needs to be redundantly configured.
  - (4) above special cases analyze in a way that makes sense if you are already familiar with GDT but otherwise is a bit confusing.
  - More significant breaking change than others we might want.

grayside:
Closing old issues that I do not expect to be tackled in the context of this project.

Issue #30, opened 2014-10-27
Use "minimal Gruntconfig" as implicit defaults


grayside:
Rather than universally mandate basic configuration, let's handle that implicitly and come as close to zero-configuration as possible. All we need to do is pull srcPaths and buildPaths into the codebase as fallback settings.

This will require adjustments to the documentation.

grayside:
Closing old issues that I do not expect to be tackled in the context of this project.

PR #285, opened 2016-06-07
Making binary paths configurable with defaults

### LINKED ISSUE #102: Add option for dependency paths

arithmetric:
- Exposing binary paths (phpcs, phpmd, drush, behat) as configuration options in Gruntconfig.json.
- Adding default values for Composer-installed dependencies.

Resolves #102. Also relates to #169 and #30.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Dropped support for Node.js v0.12 and earlier and PHP v5.4 and earlier.
 - Removed built-in support for Compass theme compilation and Ruby bundle
 installation.
+- Added configuration for the Behat binary path. Add default configuration for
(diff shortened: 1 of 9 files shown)


grayside:
Shouldn't phpcs and the tools using `path` be switched to `cmd` as well?

arithmetric:
@grayside Not sure if I understand. I tried to make all of the tools consistent in using `path`.

grayside:
Ah, my brain flipped those changes. Looks good. Something we'll want to add is some clearer docs on how to turn off things by setting them to `false` or removing from the Gruntconfig. That's not a blocker here though.

PR #286, opened 2016-06-09
Validate YAML files by default

jhedstrom:
The Drupal coder module has some basic standards around `yml` files. It doesn't actually lint them for parsing errors, but it enforces the new line rule, and also warns folks to remove the `version` line if it exists (which console's `drupal generate:module` likes to add `version: VERSION` by default for some reason).


grayside:
Great little addition. To add patterns just for PHPCS, and not eslint, PHPLint, or PHPMD, this should be added around line 53. If the other tools don't have bugs/performance regressions/etc trying to process it where it is, that's fine.

In reviewing some travis tests, it looks fine to me, but could you check this with the `grunt analyze` command to make sure of PHPMD?

jhedstrom:
I tried moving this down to line 53:

```
phpcsPatterns.unshift('<%= config.srcPaths.drupal %>/{modules,profiles,themes,static}/**/*.{yml}');
```
(both via `push` and `unshift`), but it didn't seem to pick up yml files. Testing with `grunt analyze` didn't seem to like yaml files for all the other checks.

arithmetric:
@jhedstrom Do you mean you were seeing an error with `grunt analyze`?

Trying to see where this PR stands. I don't see any issues.

jhedstrom:
@arithmetric grunt didn't seem to like having 2 different patterns (it works with one inclusive, and multiple exclusionary (`!`), but multiple inclusionary wasn't working--it seems to only pick).

arithmetric:
@jhedstrom I took a look at this one, and it seems that the curly brackets are causing an issue when there is only one item (curly brackets are used to indicate a set of items).

This worked on its own around line 53:

```
phpcsPatterns.unshift('<%= config.srcPaths.drupal %>/{modules,profiles,libraries,static}/**/*.yml');
```

I'll push an update to this branch with that change and a merge from master.

Issue #289, opened 2016-06-21
Remove Composer dev dependencies from package output


arithmetric:
When a Composer based build is packaged, development dependencies in `composer.json` should be excluded from the vendor resources included in the package output.

Follow up from #284.

jhedstrom:
We should also regenerate the autoload files so they don't include the dev dependencies. This may be automatically done depending on how we remove them (otherwise `composer dump-autoload --no-dev` would do it).

grayside:
If we end up running dump-autoload for a package, should we also run with `--optimize`?

This still leads to pushing the dev dependencies, I found `composer remove --update-no-dev` as a possibility, though I haven't tested to see if that's frustratingly slow.

jhedstrom:
In manual testing, running `composer install --no-dev` works, and is super fast. We could run this *after* packaging so as not to change the local install.

jhedstrom:
This is merged now.

PR #290, opened 2016-06-22
Rebuild package without dev dependencies

jhedstrom:
- Addresses #289

### DIFF
--- a/tasks/package.js
+++ b/tasks/package.js
@@ -50,6 +50,15 @@ module.exports = function(grunt) {
     tasks.push('clean:packages');
     tasks.push('copy:package');
 
+    // Rebuild composer dependencies without dev.
+    grunt.config(['composer'], {
+      options: {
+        flags: ['no-dev'],
+        cwd: destPath
+      }
+    });
+    tasks.push('composer:install');
+
     if (this.args[0] && this.args[0] === 'compress') {
       grunt.loadNpmTasks('grunt-contrib-compress');
       grunt.config('compress.package', {
--- a/tasks/package.js
+++ b/tasks/package.js
@@ -15,6 +15,7 @@ module.exports = function(grunt) {
     var config = grunt.config.get('config.packages');
     var srcFiles = ['**', '!**/.gitkeep'].concat((config && config.srcFiles && config.srcFiles.length) ? config.srcFiles : '**');
     var projFiles = (config && config.projFiles && config.projFiles.length) ? config.projFiles : [];
+    console.log(projFiles);
 
     // Look for a package target spec, build destination path.
     var packageName = grunt.option('name') || config.name || 'package';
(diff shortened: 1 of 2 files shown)


tasks/package.js
```diff
@@ -50,6 +50,15 @@ module.exports = function(grunt) {
...
     tasks.push('copy:package');
 
+    // Rebuild composer dependencies without dev.
+    grunt.config(['composer'], {
+      options: {
+        flags: ['no-dev'],
+        cwd: destPath
+      }
+    });
+    tasks.push('composer:install');
```
grayside: Can anyone think of a reason someone would create a task that executes package followed by a composer operation that needs -dev?

arithmetric: @grayside That would only be a problem if also executed in the packaged output, right? The dev dependencies will still exist in the project root.

arithmetric:
It looks like the tests are failing for D7. Can this Composer task be conditional on if the Composer build process is used?

tasks/package.js
```diff
@@ -50,14 +51,18 @@ module.exports = function(grunt) {
...
-    grunt.config(['composer'], {
-      options: {
-        flags: ['no-dev'],
-        cwd: destPath
-      }
-    });
-    tasks.push('composer:install');
+    // If the `composer.json` file is being packaged, rebuild composer dependencies without dev.
+    if (projFiles.find(function (pattern) {
+      return pattern.startsWith('composer');
```
grayside: Is this recursive? Is it a problem if it finds composer.lock but composer.json is missing?

jhedstrom: It shouldn't be recursive, it's just looking at the actual `projFiles` array specification (`composer.*` is the pattern in the test config).

grayside:
The latest commit raises a question to my mind:

If Drupal core has a composer.json and the project has it's own composer.json... how do we put both into the Drupal docroot and run composer install? Shouldn't that recurse reality?

jhedstrom:
> how do we put both into the Drupal docroot and run composer install? Shouldn't that recurse reality?

This expects the following packaged structure:

```
\
composer.json
composer.lock
README.md
...
build\
  html\
  index.php
  ...
  core\
    composer.json

Note there is no `composer.json` file at the docroot, only in `core`. This differs from downloading a tarball of core, or using the core git repo directly since those *do* include a docroot-level `composer.json` file. The recursion there is handled by `wikimedia/composer-merge-plugin`, which then descends into the core directory.

grayside:
Does the `composer install --no-dev` re-download already downloaded dependencies?

jhedstrom:
Not once there is a `.lock` file. In manual testing it takes a matter of seconds since the libraries are cached by composer.

jhedstrom:
screen shot 2016-06-23 at 3 31 31 pm

Issue #292, opened 2016-07-06
Composer Scaffolding fails when a composer.lock file exists


ndrake86:
In GDT 1.0-alpha if a composer.lock file exists the grunt task will execute composer:install and attempt to run the composer:drupal-scaffold task. When that process runs the following error is thrown. 

```
Running "composer:drupal-scaffold" (composer) task
> DrupalComposer\DrupalScaffold\Plugin::scaffold
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Composer\Installer\InstallationManager::getInstallPath() must implement interface Composer\Package\PackageInterface, null given, called in /var/www/vendor/drupal-composer/drupal-scaffold/src/Handler.php on line 205 and defined in phar:///usr/local/bin/composer/src/Composer/Installer/InstallationManager.php:227
Stack trace:
#0 /var/www/vendor/drupal-composer/drupal-scaffold/src/Handler.php(205): Composer\Installer\InstallationManager->getInstallPath(NULL)
#1 /var/www/vendor/drupal-composer/drupal-scaffold/src/Handler.php(102): DrupalComposer\DrupalScaffold\Handler->getWebRoot()
#2 /var/www/vendor/drupal-composer/drupal-scaffold/src/Plugin.php(77): DrupalComposer\DrupalScaffold\Handler->downloadScaffold()
#3 phar:///usr/local/bin/composer/src/Composer/EventDispatcher/EventDispatcher.php(255): DrupalComposer\DrupalScaffold\Plugin::scaffold(Object(Composer\Script\Event))
#4 phar:///usr/local/bin/composer/src/Composer/EventDispatcher/EventDispatcher.php(209): Composer\Event in phar:///usr/local/bin/composer/src/Composer/Installer/InstallationManager.php on line 227
```

Removing the lock file so that the process runs the composer:update task does not cause the scaffold task to fail.

jhedstrom:
hmm, we need the scaffold task to run after install, because otherwise none of those files in the docroot would be downloaded for devs that are running once there is a `composer.lock` file. We also don't want to run `composer update` automatically except for the first install. Running that to update the lock file should always be intentional. I'll see if I can figure out what's going on with that error.

ndrake86:
Yes this can be closed.

ndrake86:
This was a bad configuration on my part @jhedstrom thanks for the help closing out.

PR #293, opened 2016-07-06
Remove unneeded call to drupal-scaffold.

ndrake86:
There is an unneeded call to composer:drupal-scaffold in the bootstrap.js file. Scaffold is called regardless of if a composer install or composer update is called.

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -60,7 +60,6 @@ module.exports = function(grunt) {
   tasksDefault.push('scaffold');
 
   if (grunt.file.exists('./composer.lock') && grunt.config.get(['composer', 'install'])) {
-    // Manually run `composer drupal-scaffold` since this is only automatically run on update.
     // Run `composer install` if there is already a lock file. Updates should be explicit once this file exists.
     tasksDefault.unshift('composer:install');
   } else if (grunt.config.get(['composer', 'update'])) {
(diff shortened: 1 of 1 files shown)


jhedstrom:
This addresses #292.

bootstrap.js
```diff
@@ -61,7 +61,6 @@ module.exports = function(grunt) {
 
   if (grunt.file.exists('./composer.lock') && grunt.config.get(['composer', 'install'])) {
     // Manually run `composer drupal-scaffold` since this is only automatically run on update.
```
jhedstrom: The comment here should be removed too :)

jhedstrom:
I have manually tested this, and indeed the extra `composer drupal-scaffold` call is not needed when `composer install` is run. I know manually testing this weeks ago that this was not the case.

The test fails appear unrelated.

jhedstrom:
Actually, this isn't the case. The above tests were done using the `drupal/drupal` package instead of the `drupal/core` package. We still need to manually run `composer drupal-scaffold` when the `drupal/core` package is used.

arithmetric:
@jhedstrom Based on your findings, this PR should be closed without merging, right? My understanding is that this issue stemmed from using the wrong Composer package for Drupal core.

jhedstrom:
@arithmetric that's correct. Closing now.

PR #294, opened 2016-07-14
Conditionally run composer:drupal-scaffold.

jhedstrom:
- Only run the `composer:drupal-scaffold` task if it is defined in the
  `composer.json` scripts section.

This fixes an issue where D7 builds (or D8  builds without the `drupal-scaffold` script defined) would fail with:

```
  [Symfony\Component\Console\Exception\CommandNotFoundException]
  Command "drupal-scaffold" is not defined.


Warning: Task "composer:drupal-scaffold" failed. Use --force to continue.

Aborted due to warnings.
```


arithmetric:
Looks good, thanks!

Issue #288, opened 2016-06-21
Allow changing vendor path for package output


arithmetric:
When a Composer based build is packaged, the paths in the autoloader files generated by Composer should be updated to match the structure of the package output.

The goals are:

- It should be possible to created package output that can live in the docroot without references outside the docroot.

- It should also be possible to configure the vendor directory to live at the same level as the docroot.

Follow up from #284.

arithmetric:
I created this branch with comments on a possible approach to this issue:
https://github.com/phase2/grunt-drupal-tasks/compare/feature/package_rewrite_composer

I'd welcome feedback on the best approach for this. @jhedstrom @grayside @mike-potter

grayside:
I missed the last update here, will check it out soon.

In the meantime, Pantheon may have simplified the problem just recently: https://pantheon.io/docs/nested-docroot/

grayside:
In review of the proposed approach I think it's on the right track. If the GDT buildpath does not match the configured package docroot it is set up to rewrite the composer.json before re-running the `composer install` we put in place to remove dev dependencies.

I have potentially breaking changes in mind with #255 and #24 but nothing I see here is out of the way for those changes to similarly support.

mike-potter:
Created a PR for this here: https://github.com/phase2/grunt-drupal-tasks/pull/297

Closing this issue since we can discuss it now in the PR.

Issue #271, opened 2016-04-10
Write 1.0 upgrade guide


grayside:
- implementing compass in your theme
- config key changes

mike-potter:
New mkdocs support is in https://github.com/phase2/grunt-drupal-tasks/pull/300.  This should contain the config key changes.  For the compass docs, deferring this issue till after 1.0.

PR #297, opened 2016-09-22
Allow changing vendor path for package output

mike-potter:
I implemented the approach suggested by arithmetic.  Comments welcome.

### DIFF
--- a/example/Gruntconfig.json
+++ b/example/Gruntconfig.json
@@ -11,7 +11,7 @@
     "srcFiles": ["!sites/*/files/**", "!xmlrpc.php", "!modules/php/*"],
     "projFiles": ["README*", "bin/**", "hooks/**", "src/*.make", "vendor/**"],
     "dest": {
-      "docroot": "build/html",
+      "docroot": "html",
       "devResources": ""
     }
   },
--- a/tasks/package.js
+++ b/tasks/package.js
@@ -19,14 +19,17 @@ module.exports = function(grunt) {
       // Load `composer.json` as JSON, convert to object.
       var composer = grunt.file.readJSON('composer.json');
       for (var key in composer.extra['installer-paths']) {
-        var newKey = key.replace(regex, pathPackage);
-        if (newKey != key) {
-          // Alter keys in `extra.installer-paths` object to change `build/html`
-          // to `html` or an alternative path from the config.
-          var value = composer.extra['installer-paths'][key];
-          delete composer.extra['installer-paths'][key];
-          composer.extra['installer-paths'][newKey] = value;
-          changed = true;
+        // Add an unnecessary if check just for eslint rules.
+        if (composer.extra['installer-paths'][key]) {
+          var newKey = key.replace(regex, pathPackage);
+          if (newKey !== key) {
+            // Alter keys in `extra.installer-paths` object to change `build/html`
+            // to `html` or an alternative path from the config.
+            var value = composer.extra['installer-paths'][key];
+            delete composer.extra['installer-paths'][key];
+            composer.extra['installer-paths'][newKey] = value;
+            changed = true;
+          }
         }
       }
       if (changed) {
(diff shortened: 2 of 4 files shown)


mike-potter:
Referenced from this issue: https://github.com/phase2/grunt-drupal-tasks/issues/288

mike-potter:
This still needs additional work though.  It covers the build paths in the composer.json but it doesn't cover the case of moving the vendor directory and updating the autoload.php.

mike-potter:
Yeah, needs a lot more work.  When testing, the composer.json that ended up in the package was still the original.  The current PR modifies the composer.json in the root instead of the one being packaged.

mike-potter:
OK, updated this PR.  Added a new Gruntconfig.json option:

packages : dest : "vendor"

You can set this to the desired location of the /vendor directory.  If left out or empty the composer.json will be saved to the normal buildPaths.html.  If "vendor" is used, it is a relative path to the buildPaths.html that specifies where to place the composer.json file.

The installer-paths in the composer.json are updated based on the settings of both the packages.dest.docroot as well as packages.dest.vendor.  When the vendor is set, the installer-paths need to be made relative to the location of composer.

So, a simple test case:

"docroot": "docroot",
"vendor": "docroot"

This will put the composer.json within packages/package/docroot which is where the vendor directory will get created, and the installer-paths generate from that same directory.  This would be the typical setup for Acquia.

Another contrived example:

"docroot": "build/html",
"vendor": "build"

This puts composer.json into the packages/package/build folder, but sets the installer-paths to "html/*".  This would let the web server point to /build/html but have the vendor directory in /build which might still be within the hosting file structure.

arithmetric:
@mike-potter Thanks for taking this up!

The test cases you listed are good ones, but I'd simplify them slightly:

1. Everything ends up in `build/packages/package` (all output from build/html, the vendor directory, and the composer files). This is most like the default packaging output from previous versions of Grunt Drupal Tasks. I would expect to get this output by leaving `packages.dest` undefined/empty in Gruntconfig.json or setting it explicitly to:

```
"packages": {
  "dest": {
    "docroot": "",
    "vendor": ""
  }
}
```

2. The likely best approach is to have the vendor directory live in the same directory as the docroot directory (one level up from index.php). In this case, the `build/packages/package` directory would include the docroot in `html`, the vendor directory, and the composer files. I would expect to get this output with the following configuration:

```
"packages": {
  "dest": {
    "docroot": "html"
  }
}
```

When testing these two configurations, I found a few issues:

- The updated code treats `config.packages.dest.docroot` as a required setting, where it used to be optional and only needed when changing from the default of `build/packages/package`. Currently if you do not define `config.packages.dest.docroot`, the composer.json install-paths are updated with an `undefined/` path prefix.

- This solution does not update the path in the `autoload.php` in the docroot (which is created  by the drupal-scaffold package during build). One way I found to update this path is to run the Composer `drupal-scaffold` script in the package directory after other changes. However this script also triggers the full scaffold step, including downloading core and other components. Ideally we would only run the generateAutoload routine: https://github.com/drupal-composer/drupal-scaffold/blob/master/src/Handler.php#L133

I'll submit a PR against this branch with updates that I'm testing to resolve these issues.

mike-potter:
There we go...travis is passing now.  Seems to be a bit flaky these days and just re-running the builds worked.  Adam is out for 3 weeks so I'm going to ping Jonathan for a review/test and then we can commit this.  I'll be working on some docs but hope we can do the 1.0 milestone very soon.

jhedstrom:
This looks good to me. @arithmetric any further feedback here?

arithmetric:
I think this is ready to go. I'll push out another alpha release with this.

Let's make sure documentation is up-to-date (address #271) and aim to release 1.0.0 in the next couple weeks.

mike-potter:
I'm going to work on the documentation today and later this week.  I really want to push the 1.0.0 release asap because we have D8 projects that need the composer support and the current projects using gdt "^0.11.0" are not seeing any of these "alpha" releases, even though there is nothing in them that would break existing drush-make-based projects.

PR #298, opened 2016-09-24
Updates to feature/package_rewrite_composer

arithmetric:
Updates to feature/package_rewrite_composer:
- Regenerates Drupal docroot autoload.php if needed
- Allow packages.dest configuration to be optional
- Changes example and test configuration's package docroot dest to "html" so package directory should include "html" (docroot), vendor, composer.json, composer.lock

@mike-potter Please review these changes against #297.


mike-potter:
Thanks for the help on this!  I hadn't realized "docroot" was optional and your changes all make good sense.

PR #300, opened 2016-09-28
Feature/mkdocs

mike-potter:
This updates the documentation to support mkdocs, fixes long line wrapping, and cleans up the links.


arithmetric:
Thanks for pushing through on these doc updates. I'm going to merge this.

For clarity, are both methods for building docs (`build_docs.sh` and `npm run docs`) needed? Are they for different purposes?

mike-potter:
The `npm run docs` uses the preferred docker method for doing the build on systems that do not have mkdocs installed.  The `build_docs.sh` is for building the docs on a system that already has mkdocs installed.  I ran into this because I was doing dev within a docker container and couldn't run the docker-compose because that would be doing docker within docker.

Thanks for merging!

PR #301, opened 2016-10-04
Add options to package task to exclude paths such as node_modules and…

mike-potter:
Several projects using Pattern lab are having bad performance issues when using the "grunt package" command.  This PR adds an option "config.packages.exclude" (default TRUE) that will exclude additional folders specified in "config.packages.excludePaths".  If "config.packages.excludePaths" is empty, a default of ["node_modules", "bower_components"] is used.

While the existing "config.packages.srcFiles" and "config.packages.projFiles" allow you to add specific files and paths to be excluded, the "config.packages.excludePaths" is designed for folders and is applied to both srcFiles and projFiles and has the useful default paths.

A further performance improvement would be to use rsync instead of grunt-contrib-copy, but delaying that till post 1.0 release.

### DIFF
--- a/tasks/package.js
+++ b/tasks/package.js
@@ -69,12 +69,12 @@ module.exports = function(grunt) {
     var exclude = grunt.config('config.packages.exclude');
     if (exclude !== false) {
       var excludePaths = grunt.config('config.packages.excludePaths');
-      if (!excludePaths || excludePaths.length == 0) {
+      if (!excludePaths || !excludePaths) {
         excludePaths = ['bower_components', 'node_modules'];
       }
 
-      for (var key in excludePaths) {
-        excludePaths[key] = '!**/' + excludePaths[key] + '/**';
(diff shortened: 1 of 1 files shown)


mike-potter:
I commented on that in the initial PR message:  the new excludePaths applies to both the srcFiles *AND* the projFiles and provides a default option so people don't have to specifically remember to exclude these common node_modules and bower_component paths, but can override them if for some reason they want/need these files in their package.

We want to support this performance increase without everyone needing to update their Gruntconfig file.

grayside:
Sorry for missing that. Not sure we want two ways to do the exclusion, and we can certainly merge these two into any configuration. Flipside, this is closer to the syntax for an rsync based operation.

Switching grunt package up will probably provoke a 2.0 release, so I suppose this works for now.

mike-potter:
Closing this in favor of a new PR that does rsync and has a significant performance boost.

PR #302, opened 2016-10-04
Exclude node_modules and bower_components from phpmd

mike-potter:
Continuing performance issues with node_modules and bower_components, this PR adds exclusions to the phpmd command and also adds a config option to extend it.


grayside:
Depending on the semantics of exclude, might it be more performant to build the excludes from configured themes?

mike-potter:
From my research the phpmd pattern checking is very basic.  It's not even regex, its just like a "string contains".  So this PR should eliminate any node_modules or bower_modules anywhere they might be and should be as performant as anything else.  Let's keep it simple unless it doesn't work.

PR #303, opened 2016-10-05
Add rsync option to package command and improve performance

mike-potter:
This is a fairly significant patch, but should not affect existing projects.  It adds a "packages.rsync" option (False by default).  When set to true, the package will be generated in a separate folder (packages.vmData defaults to "build/vm_data") via rsync and the compress option is enabled to generate an archive from the package, which is moved to the normal build/packages destination.

By mounting build/vm_data into the VM in your docker_composer.yml file, this allows the package to be built directly to the VM bypassing NFS to the local filesystem.  Only the final archive.tgz is moved into the local file system.  The vmData persists so the Build container can still perform additional tasks as needed, such as pushing the package into a remote repo.  Or, the local tgz archive can be used to deploy the package.

On a current real project, this reduced the normal "grunt package" command running from the build container from a 40-minute task, down to a TWO minute task.

Given the significant performance improvement, I'm marking this PR for Milestone 1, so let's get reviews of this approach.  (FYI it was modeled from some code in a "grunt deploy" task written for another project by Adam)

### DIFF
--- a/tasks/package.js
+++ b/tasks/package.js
@@ -61,10 +61,17 @@ module.exports = function(grunt) {
 
   grunt.registerTask('package', 'Package the operational codebase for deployment. Use package:compress to create an archive.', function() {
     grunt.loadNpmTasks('grunt-contrib-copy');
+    grunt.loadNpmTasks('grunt-shell');
 
     var config = grunt.config.get('config.packages');
     var srcFiles = ['**', '!**/.gitkeep'].concat((config && config.srcFiles && config.srcFiles.length) ? config.srcFiles : '**');
     var projFiles = (config && config.projFiles && config.projFiles.length) ? config.projFiles : [];
+
+    // Look for a package target spec, build destination path.
+    var packageName = grunt.option('name') || config.name || 'package';
+    var destPath = grunt.config.get('config.buildPaths.packages') + '/' + packageName;
+    var tasks = [];
+    grunt.option('package-dest', destPath);
 
     var exclude = grunt.config('config.packages.exclude');
     if (exclude !== false) {
@@ -76,44 +83,50 @@ module.exports = function(grunt) {
       for (var i = 0; i < excludePaths.length; i++) {
         excludePaths[i] = '!**/' + excludePaths[i] + '/**';
       }
-      srcFiles = srcFiles.concat(excludePaths);
-      projFiles = projFiles.concat(excludePaths);
+      //srcFiles = srcFiles.concat(excludePaths);
+      //projFiles = projFiles.concat(excludePaths);
     }
 
-    // Look for a package target spec, build destination path.
-    var packageName = grunt.option('name') || config.name || 'package';
-    var destPath = grunt.config.get('config.buildPaths.packages') + '/' + packageName;
-    var tasks = [];
-    grunt.option('package-dest', destPath);
+    var rsync = 'rsync -a '
+        + '--exclude node_modules '
+        + '--exclude bower_components '
+        + '--exclude .git '
+        + '--exclude sites/*/files '
+        + '--exclude xmlrpc.php ';
 
-    grunt.config('copy.package', {
-      files: [
-        {
-          expand: true,
-          cwd: '<%= config.buildPaths.html %>',
-          src: srcFiles,
-          dest: path.resolve(destPath, grunt.config.get('config.packages.dest.docroot') || ''),
-          dot: true,
-          follow: true
-        },
-        {
-          expand: true,
-          src: projFiles,
-          dest: path.resolve(destPath, grunt.config.get('config.packages.dest.devResources') || ''),
-          dot: true,
-          follow: true
-        }
-      ],
-      options: {
-        gruntLogHeader: false,
-        mode: true
-      }
+    var srcPath = grunt.config('config.buildPaths.html');
+    var destSrc = path.resolve(destPath, grunt.config.get('config.packages.dest.docroot') || '');
+    var command = rsync + srcPath + '/ ' + destSrc + '/';
+    console.log(command);
+    console.log(destSrc)
+    grunt.config('shell.mkSrc', {
+      command: 'mkdir -p ' + destSrc
     });
+    grunt.config('shell.srcFiles', {
+      command: command
+    });
+
+    var destProj = path.resolve(destPath, grunt.config.get('config.packages.dest.devResources') || '');
+    grunt.config('shell.mkProj', {
+      command: 'mkdir -p ' + destProj
+    });
+    for (var i = 0; i < projFiles.length; i++) {
+      var command = rsync + srcPath + '/' + projFiles[i] + ' ' + destProj;
+      console.log(command);
+      grunt.config('shell.projFiles' + i, {
+        command: command
+      });
+    }
 
     grunt.config.set('clean.packages', [destPath]);
 
     tasks.push('clean:packages');
-    tasks.push('copy:package');
+    tasks.push('shell:mkSrc');
+    tasks.push('shell:srcFiles');
+    tasks.push('shell:mkProj');
+    for (var i = 0; i < projFiles.length; i++) {
+      tasks.push('shell:projFiles' + i);
+    }
 
     // If the `composer.json` file is being packaged, rebuild composer dependencies without dev.
     if (projFiles.find(function(pattern) {
--- a/tasks/package.js
+++ b/tasks/package.js
@@ -61,17 +61,10 @@ module.exports = function(grunt) {
 
   grunt.registerTask('package', 'Package the operational codebase for deployment. Use package:compress to create an archive.', function() {
     grunt.loadNpmTasks('grunt-contrib-copy');
-    grunt.loadNpmTasks('grunt-shell');
 
     var config = grunt.config.get('config.packages');
     var srcFiles = ['**', '!**/.gitkeep'].concat((config && config.srcFiles && config.srcFiles.length) ? config.srcFiles : '**');
     var projFiles = (config && config.projFiles && config.projFiles.length) ? config.projFiles : [];
-
-    // Look for a package target spec, build destination path.
-    var packageName = grunt.option('name') || config.name || 'package';
-    var destPath = grunt.config.get('config.buildPaths.packages') + '/' + packageName;
-    var tasks = [];
-    grunt.option('package-dest', destPath);
 
     var exclude = grunt.config('config.packages.exclude');
     if (exclude !== false) {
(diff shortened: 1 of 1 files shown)


tasks/package.js
```diff
@@ -61,29 +61,101 @@ module.exports = function(grunt) {
...
     var packageName = grunt.option('name') || config.name || 'package';
-    var destPath = grunt.config.get('config.buildPaths.packages') + '/' + packageName;
+
+    // Determine if we are using rsync and vmPath and tarball for performance.
+    var useRsync = grunt.config('config.packages.rsync') || false;
+    // Determine if we are creating a tarball archive.
+    var archive = grunt.config('config.packages.archive') || useRsync;
+
+    // When using rsync, generate to a path that can be mounted in the VM.
+    var vmPath = grunt.config.get('config.packages.vmData') || 'build/vm_data';
```
grayside: Per earlier comments, I'd like a more contextual, less tech-specific name for this -- intermediary?

tasks/package.js
```diff
@@ -61,29 +61,101 @@ module.exports = function(grunt) {
...
+    if (destPath !== finalPath) {
+      cleanPaths.push(finalPath);
+    }
     grunt.option('package-dest', destPath);
+    grunt.config.set('clean.packages', cleanPaths);
+    tasks.push('clean:packages');
+
+    var excludePaths = ['bower_components', 'node_modules', '.gitkeep'];
+
+    if (useRsync) {
```
grayside: As previously explored with https://github.com/phase2/grunt-drupal-tasks/blob/master/tasks/make.js#L71, we need a fallback for Windows which so far has been simply "Got Windows? Okay, we'll still use copy!" Because the configuration for copy and rsync are not fully compatible, I've been thinking of rsync not as a mode, but *rsync to copy fallback* as a breaking change so we can revise configuration requirements. Open to discussion on this point.

docs/60_PACKAGE.md
```diff
@@ -51,6 +51,35 @@ directory.  Defaults to the docroot.
...
+
+**rsync**: If set to true, rsync will be used instead of file copying for
+improved performance. This automatically enables the `compress` option and will
+write the archive to the package destination.
+
+**vmData**: This path is used as the intermediate generation directory for the
+package when using the `rsync` option. By selecting a volume mounted in your
+VM `/data` area, this can significantly improve performance. The archive is
+still moved to the normal package destination when complete.  Defaults to
+`build/vm_data`.
```
mike-potter: This is the super important part of this PR.  Without the virtual intermediate directory there was no way to get anything performant and people are bypassing gdt on their projects because of this.  I know GDT is not "targetting" at containers, which is why this option isn't enabled by default.

The idea here is that in your docker container you mount a volume that gdt can use for various purposes to improve performance.  Only tackling the packaging right now, but I can imagine other commands using this also.  So don't want it to be within "build/package".  The /data is just an example that pertains to our own container, but that isn't anywhere in the GDT except the docs as an example.  Since our docs already reference many P2 practices, I don't see this as a problem right now (see the CI docs for example).

Still, specific updates to this doc file are welcome if you want to wordsmith it a bit.

docs/60_PACKAGE.md
```diff
@@ -51,6 +51,35 @@ directory.  Defaults to the docroot.
...
+
+## Performance
+
+When running `grunt package` within a docker container, the default file
```
mike-potter: I think I covered this above, but the whole point of this PR was performance.  Switching from copy to rsync gives a small performance improvement.  The majority of the improvement comes from using the vm_data.

Yes, it's really about NFS and not docker, so the docs can be wordsmithed a bit.

And yes, this vm_data is definitely something we want to incorporate into other areas of gdt.  The package command was just the big target because it was taking 40 minutes on a project.

tasks/package.js
mike-potter: Specific suggestions are welcome.

tasks/package.js
mike-potter: The code I added makes them compatible.  You still use srcFiles, projFiles, and the new excludePaths.  The code builds the proper arguments for either copy or rsync.  So yes, Windows cannot use the "rsync=true" option.

If there is a simple way to detect if rsync is installed I'd be happy to use that to set the default and have it fallback.

mike-potter:
FYI: Travis tests are failing because of https://github.com/phase2/grunt-drupal-tasks/issues/304 and not because of anything in this PR.

tasks/package.js
```diff
@@ -61,29 +61,101 @@ module.exports = function(grunt) {
...
+      // Pull negate conditions from srcFiles into excludePaths.
+      for (var srcIndex = 0; srcIndex < srcFiles.length; srcIndex++) {
+        var item = srcFiles[srcIndex];
+        if (item.substr(0, 1) === '!') {
+          item = item.slice(1);
+          item = item.replace('**', '*');
+          excludePaths.push(item);
```
grayside: This is really smart, however, the filesystem matching by the Node minimatch library has more breadth of syntax than is typically supported by Bash/Rsync/etc beyond just negation, such as fancy globbing and simple expressions. That's the incompatibility I alluded to earlier. I think we need to confirm that failure is reasonably graceful as a minimum.

tasks/package.js
```diff
@@ -117,25 +192,31 @@ module.exports = function(grunt) {
...
             src: ['**']
           }
         ]
       });
 
       tasks.push('compress:package');
+
+      if (destPath !== finalPath) {
+        grunt.config('shell.mvArchive', {
+          command: 'mv ' + destPath + '.tgz ' + finalPath + '.tgz'
```
grayside: Of note, mv will not work across NFS volume mounts. We will want to document that limitation a bit to help inform how projects are configured.

tasks/package.js
```diff
@@ -96,9 +168,12 @@ module.exports = function(grunt) {
...
-    grunt.config.set('clean.packages', [destPath]);
-
-    tasks.push('clean:packages');
+    if (useRsync) {
+      tasks.push('mkdir:package');
+      tasks.push('shell:rsync');
+    } else {
+      tasks.push('copy:source');
+    }
     tasks.push('copy:package');
```
grayside: copy:package looks like it's been stripped down to some final copy operations of random project files. Maybe this should be renamed for clarity? `copy:package-extras`?

tasks/package.js
```diff
@@ -61,29 +61,101 @@ module.exports = function(grunt) {
...
+          item = item.slice(1);
+          item = item.replace('**', '*');
+          excludePaths.push(item);
+        }
+      }
+
+      // Setup default rsync command and options.
+      var rsync = 'rsync -ahWL --no-l --stats --chmod=Du+rwx ';
+      for (var pathIndex = 0; pathIndex < excludePaths.length; pathIndex++) {
+        rsync = rsync + "--exclude " + excludePaths[pathIndex] + ' ';
```
grayside: Seems like only the final excluded file would be included here. Shouldn't we remove the loop and replace with an array join on whitespace?

tasks/package.js
```diff
@@ -61,29 +61,101 @@ module.exports = function(grunt) {
...
+
+      grunt.config('shell.rsync', {
+        command: rsyncCommand
+      });
+    } else {
+      // Use slower copy if rsync is disabled in options.
+      srcFiles.unshift('**');
+      // Add any additional excludePaths to srcFiles.
+      for (var i = 0; i < excludePaths.length; i++) {
+        srcFiles.push('!**/' + excludePaths[i]);
```
grayside: Array merge seems more clean here. Also, something to dedup might be worthwhile to make sure any code that wants to remove exclusions only has to find a single instance of a mask.

.travis.yml
```diff
@@ -44,15 +44,15 @@ script:
   - echo "Working copy disk usage"; du -chs .
   - grunt --version
   - grunt help
-  - grunt --quiet --timer
```
grayside: Was --quiet suppressing errors too? In our case it's used to suppress desktop notifications which are irrelevant from inside travis.

docs/60_PACKAGE.md
grayside: I want to make sure all config options and documentation are not overly bound to containers or virtual environments specifically, even if the changes we have facilitate that.

tasks/package.js
grayside: We have a canRsync utility function used by the code I linked to. I'd like that to be checked as part of the condition whether we are using Rsync. We can make that function smarter to detect if rsync is installed in a follow-up.

docs/60_PACKAGE.md
```diff
@@ -51,6 +51,35 @@ directory.  Defaults to the docroot.
 packages directory. This can be overridden by calling grunt package with the
 `--name` parameter.
 
+**archive**: Used to enable the `compress` option via the configuration instead
+of the command line. Defaults to false.
+
+**rsync**: If set to true, rsync will be used instead of file copying for
+improved performance. This automatically enables the `compress` option and will
+write the archive to the package destination.
```
grayside: Suppose a project leaves it creating an archive, and turns on rsync. What is the use case we're aiming for? Operational code backups? I'm open to calling this a best default, I'm just trying to understand what we're talking about empowering.

docs/60_PACKAGE.md
arithmetric: To make this less specific to Docker and the `/data` path set up by devtools, could this be recast as a temporary directory path (i.e. tempDir rather than vmData)? We can note in the documentation that the purpose for this path is to allow using a path that performs quickly if the project directory lives in a slow performing location (i.e. NFS).

.travis.yml
mike-potter: The package command had a bug that was not completing all of its tasks, so without seeing the task output it was difficult to debug.  Also hard to see the actual tasks being run through the cloud of all the other output generated in the travis log.  I don't consider those desktop notifications irrelevant for travis as they aided me in debugging.  And it doesn't hurt to have them, right?

tasks/package.js
mike-potter: Nice, I didn't see the canRsync.  I'll add that.

tasks/package.js
mike-potter: Yes, but I think we can address that in the documentation.  IMHO none of our config options should be tied to a specific implementation (node, rsync, etc).  So we should specify how wildcards are handled and then convert to the needed format for either Copy or Rsync.  

Right now the only issue is ** vs * and rsync seems to take the ** and still function, so I didn't need to change that.  

If people want fancy expressions then we should use some form of RegEx that would be less implementation-specific.  But that's work for later.  For now if somebody wants to use a fancy Node minimatch expression then they just can't use the new rsync option.

tasks/package.js
mike-potter: Yeah, I even thought of "copy:projFiles" to be really specific.  If we call it "package-extras" then having the config be "projFiles" might be confusing too.

tasks/package.js
mike-potter: So that's what I thought too.  But it seems to work across docker volumes.  Maybe we should change it from a "mv" to a normal grunt copy.  Then delete the tarball from the destPath.

docs/60_PACKAGE.md
grayside: Gotcha, writing the tarball being less I/O, offset by the time to create the tarball. Have you noted the time to create the tarball is less than syncing the files individually? I expect so.

tasks/package.js
grayside: Per other thread, we're calling it tempDir. Just closing the loop here :) Certainly shorter than `intermediary`.

tasks/package.js
grayside: - [ ] Confirm using a minimatch expression results in rsync failing in a reasonably informative and quick manner.
- [ ] Release notes and documentation detail the new restriction to stick with Bash-compatible file matching patterns with rsync mode.

tasks/package.js
grayside: Something like:

`var _ = require('lodash');` at the top.

```
srcFiles = _.union(srcFiles, _.map(excludePaths, function (item) {
  return '!**/' + item;
}));
```

Union both merges and unique-ifies.

* https://lodash.com/docs/4.16.4#map
* https://lodash.com/docs/4.16.4#union

Looks like we should check on updating lodash.

tasks/package.js
grayside: That seems fine. This is a stepping stone to #255, otherwise I'd suggest we drop projFiles and rename the config for clarity.

tasks/package.js
grayside: Hm, maybe that restriction is only outside the container. mv is an order of magnitude faster than copy, so if it works I think we should keep it until we have reason to switch to copy.

A bigger issue is windows compat. For portability we should just switch to an fs.rename[Sync].

.travis.yml
arithmetric: Both the `--quiet` option and `GDT_QUIET` env var are supposed to only affect desktop notifications (i.e. growl). If the option affects the error reporting from Grunt, we should change the option to avoid conflicts.

arithmetric:
@mike-potter Could we target this for a 1.1.0 milestone? I think the current master version is in good shape for a rc1 release and a stable one after a bit of testing.

mike-potter:
Sure, we can do 1.1.0.  I agree that we need to get 1.0.0 released so people can start using the composer-based support.  And adding the "tempdir" option probably opens up other issues we'd want to work on for improving performance elsewhere in GDT that could benefit.

tasks/package.js
mike-potter: This is adding "--exclude path" to rsync for each item in the array.  So maybe a join with "--exclude " would work?  Otherwise this loop does actually work fine.

docs/60_PACKAGE.md
```diff
@@ -51,6 +51,35 @@ directory.  Defaults to the docroot.
...
+be very slow. Using the `rsync` option along with creating the `build/vm_data`
+intermediate mount point in your docker container can improve performance by
+a factor of 20 or more. However, the downside is only the compressed archive of
+your package will be written to your local disk. The full package folder will
+only be accessible within your docker container.
+
+A sample docker-composer.yml entry to mount the vm_data looks like this:
+```
+  volumes:
+    - /data/PROJECTNAME/package:/var/www/build/vm_data
```
grayside: I've picked up this branch and am trying to pull it along. As part of this I'm reviewing and revising the docs to clarify my game plan. I'm confused why we would volume mount the intermediary directory. Wouldn't that slow down the rsync operation compared to leaving the tempDir somewhere in the container but **not** in a volume mount where NFS slowdowns can catch it?

mike-potter: The /data volume is just in the VM, not on the local Mac NFS system. This simply makes the stuff in /var/www/build/vm_data persistent, similar to what we do with the MySQL database. Volumes are only slow if they map the local Mac folders

Issue #304, opened 2016-10-05
Drupal 8.2 requires a /profiles directory


mike-potter:
When there isn't any src/profiles then the Drupal /profiles directory does not get created:

Running "symlink:profiles" (symlink) task
>> Created 0 symbolic links.

In Drupal 8.2, this causes an error:

PHP Fatal error:  Uncaught exception 'UnexpectedValueException' with message 'RecursiveDirectoryIterator::__construct(/home/travis/build/phase2/grunt-drupal-tasks/test/working_copy/build/html/profiles): failed to open dir: No such file or directory' in /home/travis/build/phase2/grunt-drupal-tasks/test/working_copy/build/html/core/tests/bootstrap.php:23

Currently causing our travis tests to fail.

mike-potter:
Addressing this issue in https://github.com/phase2/grunt-drupal-tasks/pull/305.

grayside:
To be clear, in Drupal 8 the top-level profiles directory doesn't exist by default? There's nothing in GDT that is responsible for creating or removing that directory to my knowledge.

mike-potter:
Correct.  In Drupal 8.1, not having a /profiles directory did not cause a problem.  I submitted https://www.drupal.org/node/2812607#comment-11698201 as a core issue for this regression in 8.2.  But no reason not to ensure the directories get created.  Nothing in drupal core is responsible for creating /profiles.  It just comes are part of the tar.gz, but not when you reference core in composer.json and not done by drupal-scaffold.  If there isn't any custom profile in /src/profiles, the /profiles dir was not created by gdt.

arithmetric:
Fixed by #305.

PR #305, opened 2016-10-05
Ensure profile, modules, libraries directories exist

mike-potter:
This is a fix for https://github.com/phase2/grunt-drupal-tasks/issues/304.  In Drupal 8.2 it requires the existence of the modules, libraries, profiles folders.  If there isn't any corresponding folder in /src, these top level folders were not getting created.

### DIFF
--- a/tasks/scaffold.js
+++ b/tasks/scaffold.js
@@ -43,7 +43,6 @@ module.exports = function(grunt) {
           src: '<%= config.srcPaths.drupal %>/modules',
           dest: path.join(drupal.modulePath(), 'custom')
         });
-        console.log("Profile: ",drupal.profilePath());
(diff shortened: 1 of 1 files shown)


grayside:
To be clear, the goal here is to make sure /modules, /libraries, and /profiles is created? I raised the question on profiles in #304, but I'm also confused why modules directory needs to be created.

mike-potter:
Yes, just making sure all the normally required directories are created.  For example, what if /src doesn't have any custom modules, so /src/modules is missing?  Similar to what is happening for /src/profiles being missing.  Just trying to cover all the required directories since it doesn't hurt to ensure they are created until Drupal is fixed to not require them.

PR #312, opened 2017-01-05
Defer to phpcs.xml file if present.

### LINKED ISSUE #311: Defer to phpcs.xml/phpcs.xml.dist when present
When a project's files grow beyond a certain point, the `grunt validate` task stops working since every single file to be validated is passed to the `phpcs` command.

The error is

```
Warning: spawn E2BIG Use --force to continue.
```

To avoid this, it seems like a good idea to defer to a `phpcs.xml` or `phpcs.xml.dist` file (`phpcs` itself will use these if no files are passed).

jhedstrom:
- Fixes #311


jhedstrom:
This probably needs some tests. Also, it may require changes to `grunt-phpcs` since even passing empty files, that module is still specifying the standard, etc:

```
Executing: vendor/bin/phpcs --report=full --standard=vendor/drupal/coder/coder_sniffer/Drupal,vendor/drupal/coder/coder_sniffer/DrupalPractice ""
```

jhedstrom:
Need to add 2 warnings here:

1. If phpcs paths/dirs are specified in `Gruntconfig.js` *and* there is a `phpcs.xml` file, need to warn that the grunt configs will be ignored.
1. If there is a `phpcs.xml` file and there is a conflicting standard specified in `Gruntconfig.js`, warn that this standard should be set to `phpcs.xml`.

jhedstrom:
@grayside I've added the first warning, but cannot remember how the second warning actually differs.

jhedstrom:
Hmm, this still needs a bit of work I think. The standard in the phpcs file isn't actually being used as the code in `quality.js` is still specifying the standard.

PR #320, opened 2017-03-03
Allow task-specific Composer arguments and apply defaults

### LINKED ISSUE #319: Default and custom Composer arguments not applied
Currently, the default Composer arguments for install and update commands are not applied to the Composer command at run-time.

This can be verified by altering `task/composer.coffee` in grunt-composer to add debugging output before the command is exec'ed. See: https://gist.github.com/arithmetric/9aa44be88d8bde383a8bbe0d7a0e1b2b

Additionally, it is not possible to re-configure the Composer arguments for a particular Composer task (i.e. only for install and update, not for drupal-scaffold).

This is due to grunt-composer being initialized in "simple task mode" instead of "multi task mode". See: https://www.npmjs.com/package/grunt-composer#simple-task-mode

arithmetric:
This fixes #319.

Changes:
- Initializes grunt-composer after configuration to enable multi task mode and task-specific configuration.
- Updates default tasks for multi task mode.

At this point, this has a potential breaking change, since the Grunt commands `composer:install` and `composer:update` no longer work and must be changed to `composer:install:install` and `composer:update:update`.

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -62,13 +62,13 @@ module.exports = function(grunt) {
   if (grunt.file.exists('./composer.lock') && grunt.config.get(['composer', 'install'])) {
     if (grunt.config.get(['composer', 'drupal-scaffold'])) {
       // Manually run `composer drupal-scaffold` since this is only automatically run on update.
-      tasksDefault.unshift('composer:drupal-scaffold');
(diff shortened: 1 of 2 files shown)


grayside:
This is surfacing bugs in the package command for Drupal 8, which we already know is badly broken. Going to play with this some more but will probably merge as a precursor to fixing `package`.

PR #323, opened 2017-08-16
Add symlink for src/config directory

mike-potter:
Added steps in scaffold.js to add the symlink from drupal docroot/config to src/config.

### DIFF
--- a/lib/drupal.js
+++ b/lib/drupal.js
@@ -51,6 +51,10 @@ module.exports = function(grunt) {
     return path.join(grunt.config('config.buildPaths.html'), 'profiles');
   };
 
+  module.configPath = function() {
+    return path.join(grunt.config('config.buildPaths.html'), 'config');
+  };
+
   module.sitePath = function() {
     return path.join(grunt.config('config.buildPaths.html'), 'sites');
   };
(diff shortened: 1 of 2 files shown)


grayside:
We do not need to do Drupal core version detection, we should instead make the presence of src/config optional.

scottalan:
Brought this up in a meeting with @grayside. Wanted to link to this issue here: https://www.drupal.org/node/2899534

scottalan:
> Basic test coverage is needed. This will require creating examples/src/config/.gitkeep

@grayside or @mike-potter  In Atmosphere we are using `config_split`. We opted to create our directory structure like this, `src/config/default`, where default holds our main site configuration. We then have configuration that only exists in specific environments in other directories. Do we think it makes sense to create `src/config/default/.gitkeep` to make it easier for projects that will need to use config sync? I ask this purely from an organizational standpoint.

grayside:
There are two basic approaches we can take as far as symlinks: we can link to src/config from build/html/config, or we can link to src/config/* from build/html/config/*. For the latter, I mean the approach taken with symlinking profiles as opposed to the approach of symlinking modules (which uses build/html/modules/custom --> ../../../../src/modules.

I do not think gdt needs to present support for Config Split, but if having a config/default directory seems reasonable in the absence of config split I don't object.

grayside:
Superceded by #330

PR #325, opened 2017-08-28
Use absolute path to HTML build directory.

### LINKED ISSUE #324: Relative path for drush status not working in drush 9
Due to this drush issue (https://github.com/drush-ops/drush/issues/2909), with drush 9, the relative path used for the `--root` arg, is not working properly, so we fail to detect the drupal major version.

One side effect of this is that modules suddenly start getting dumped into `sites/all/modules` since we fall back to D7 and earlier behavior for unknown Drupal versions.

jhedstrom:
- Fixes #324

### DIFF
--- a/lib/drupal.js
+++ b/lib/drupal.js
@@ -12,7 +12,7 @@ module.exports = function(grunt) {
     }
 
     // Using direct spawn functionality to capture stdout.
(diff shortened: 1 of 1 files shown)


jhedstrom:
This should work fine with earlier versions. I'll manually test to be sure.

jhedstrom:
Verified this works with both Drush 8 and Drush 9.

PR #328, opened 2017-10-10
Adding docker files to aid in testing

scottalan:
This is what I set up to run tests. Wanted to create a PR we could use to discuss getting this in.


grayside:
Per request in #335, I'm putting together a more minimal dockerization for local test execution and will write some light docs on developing gdt in the context of a "real" project.

PR #330, opened 2017-10-24
Add config directory

scottalan:


tasks/scaffold.js
```diff
@@ -75,7 +81,18 @@ module.exports = function(grunt) {
...
-        ]);
+        ];
+
+        // We symlink the config directory if one exists.
+        if (createConfig) {
+          tasks.push('symlink:config');
+          grunt.config(['symlink', 'config'], {
+            src: '<%= config.srcPaths.drupal %>/config',
+            dest: drupal.configPath()
+          });
```
grayside: This code works just fine, but it does read as out of order to add the task before defining it.

test/build.js
```diff
@@ -139,6 +139,17 @@ describe('grunt', function() {
         done();
       });
     });
+
+    // Ensure the build/html/config directory exists if it should.
+    it('config directory should exist if provided ', function(done) {
+      var src = fs.existsSync('src/config');
+      if (src) {
+        fs.existsSync('build/html/config', function(exists) {
+          assert.ok(exists);
```
grayside: Maybe I'm fuzzy on this, but if we have a test where the src/config doesn't exist, we should confirm the build/html/config doesn't exist. If we don't, why include the check?

grayside:
Does this deprecate #323?

tasks/scaffold.js
scottalan: ✅

test/build.js
scottalan: ✅

scottalan:
@grayside yes, this deprecates #323 

I forked the repo, vs pushing straight to this branch in origin.

test/build.js
```diff
@@ -139,6 +139,22 @@ describe('grunt', function() {
...
+      var html = fs.existsSync('build/html/config');
+      if (src && html) {
+        assert.ok(true, 'Both src/config and build/html/config exist');
+        done();
+      } else if (src && !html) {
+        assert.ok(false, 'Error: src/config exists but build/html/config does not');
+        done();
+      } else if (html && !src) {
+        assert.ok(false, 'Error: build/html/config exists but src/config does not');
+        done();
```
grayside: I appreciate the thoroughness here. Do we have Drupal 7 test cases executing that won't have the config directory? If so, that seems like it deserves a separate test case.

grayside:
Re-triggering travis tests.

test/build.js
```diff
@@ -139,6 +139,32 @@ describe('grunt', function() {
         done();
       });
     });
+
+    // Ensure the build/html/config directory exists if it should.
+    it('config directory should exist if provided ', function(done) {
+      var src = fs.existsSync('src/config');
+      var html = fs.existsSync('build/html/config');
+      if (drupalCore === '8') {
```
scottalan: I believe strict equality is fine here based on what I'm seeing here:

```
Setting environment variables from .travis.yml
$ export NVM_NODE_VERSION="4"
$ export GDT_DRUPAL_CORE="7"
```

scottalan:
@grayside added a couple checks for D8 and D7 to the tests

PR #331, opened 2017-10-31
For Drupal 8 packaging, loading composer.json from repo root.

grayside:


grayside:
Merging this now on the basis of it's completely broken as-is.

PR #332, opened 2017-11-06
Allow the use of separate files for configuration

csegarra:
Instead of using one big _Gruntconfig.js_ file, this patch allows the configuration of basic options on this file and use separate files inside one directory for the configuration of the different tasks **grunt-drupal-tasks**.

To achieve this we use load-grunt-configs [0]. This module allows you to split your Grunt task configuration objects into separate files any way you choose. With it you also can configure targets for a single task in multiple files.

On the _example_ directory we migrate the configurations of the package task to use this new format, keeping the rest of the configuration on the _Gruntconfig.js_ file.

For using this feature, it only requires to add the key **config_dir** to the **srcPaths** object.

[0] https://www.npmjs.com/package/load-grunt-configs

### DIFF
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -1,6 +1,6 @@
 module.exports = function(grunt) {
   // Initialize global configuration variables.
-  var config = grunt.file.readJSON('Gruntconfig.json');
+  var config = require('load-grunt-configs')(grunt);
   if (grunt.config.getRaw() === undefined) {
     grunt.initConfig({
       config: config
--- a/bootstrap.js
+++ b/bootstrap.js
@@ -1,12 +1,32 @@
 module.exports = function(grunt) {
   // Initialize global configuration variables.
-  var config = require('load-grunt-configs')(grunt);
+  var config = grunt.file.readJSON('Gruntconfig.json');
   if (grunt.config.getRaw() === undefined) {
     grunt.initConfig({
       config: config
     });
   } else {
     grunt.config.set('config', config);
+  }
+
+  config_dir = grunt.config.get('config.srcPaths.config_dir');
+  if (config_dir && grunt.file.isDir(config_dir)) {
+    var options = {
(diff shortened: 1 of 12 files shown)


grayside:
Hi @csegarra, thank you for your contribution! I need to think about this a bit because my planning around grunt-drupal-tasks configuration handling was more toward decoupling it from GDT itself, instead leaning on generator-gadget to put in place smart configuration loading: https://github.com/phase2/generator-gadget/issues/67

However, before that can be reached, this may be a good interim step.

A couple thoughts:

1. The src/ directory is reserved for things used by Drupal more actively, and incidentally src/config is being used in #330 for Drupal configuration management. In generator-gadget, a grunt plugin loader loads anything found in bin/grunt. Maybe we should have a bin/config? I'm still thinking through the best semantics on this. Maybe it's a top-level grunt/ directory.
2. Thank you thank you for YAML! :)
3. Even more thank yous for tests, I will review closer to confirm they are sufficiently deep.

csegarra:
Hi @grayside!
Thank you for the speed of your review and your feedback.

On our first approach (commit _4c12346543180415b85c5ea0f0414fc62f02660e_) we completely remove the _Grunconfig.js_ file and replace it with a **config** dir on the root of our projects.

For this PR we add the **srcPaths.configDir**, so the changes we made maintain some backward compatibility for those projects that already use **grunt-drupal-tasks**.

But... grunt searches the directory on the root directory of the project, not on the **src** directory... my bad ;)

If you agree, I would move the **configDir** variable out of **srcPaths**.

If it's not a _config_ directory on the project's root, I prefer a _grunt/config_ directory, rather than a _bin/config_. The _bin_ directory usually contains executable files and scripts.

By the way, with the last commit I have updated the **npm-shrinkwrap.json** to include the **load-grunt-configs**.

Regards,
Cristian

PR #333, opened 2017-11-13
 Allow using packages.dest.devResources with composer.json

csegarra:
When using a non empty **packages.dest.devResources** the task **packageRewriteComposer** fails, because it does not find the _composer.json_ file on _destPath_.

This variable is set to **grunt.config.get('config.buildPaths.packages') + '/' + packageName** on line 78 of **tasks/package.js**. And then the **packageRewriteComposer** search for composer.json on that location.

But the **copy:package** task will copy the composer.json project's file to **path.resolve(destPath, grunt.config.get('config.packages.dest.devResources') || '')**, breaking the task.

With this patch I rebuild de location of the composer.json file when running **packageRewriteComposer**.

### DIFF
--- a/tasks/package.js
+++ b/tasks/package.js
@@ -17,7 +17,8 @@ module.exports = function(grunt) {
     if ((pathBuild !== pathPackage) || pathVendor) {
       // Load `composer.json` as JSON, convert to object.
       var destPath = grunt.option('package-dest');
-      var composer = grunt.file.readJSON(destPath + '/composer.json');
+      var composerPath = path.resolve(destPath, grunt.config.get('config.packages.dest.devResources') || '');
+      var composer = grunt.file.readJSON(composerPath + '/composer.json');
       // Determine new installer-paths
       var pathInstall = pathPackage ? pathPackage + '/' : '';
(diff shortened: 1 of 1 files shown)


grayside:
My solution for what seems like part of the same problem is in https://github.com/phase2/grunt-drupal-tasks/pull/331/files.

We're currently blocked on merging PRs until the tests are fixed, I'm trying to line up time to address this.

csegarra:
Hi @grayside !

Sorry, I didn't see your PR that solve this.

I like the idea of using the original _composer.json_, so there is no obligation of include it on the **projFiles** of the package. I just think that in this case, the original composer.json sould not be removed:

`
         // Remove the original file if we moved it.
         grunt.file.delete(destPath + '/composer.json');
`

Regarding the problems with the tests, I don't know much of behat, but if you can point me to some docs, I will be glad to try to help.

Regards
Cristian

PS: I close this PR.

grayside:
I captured that bit of feedback to the other PR.

I will see if it's likely to to get this fixed soon, if not I will file an issue around fixing the test failures and provide some guidance.

PR #334, opened 2017-11-17
Correct path of the config.serve.concurrent configuration

csegarra:
The **config** prefix of the path **serve.concurrent** used on the **serve** task is missing.

Without it, there is no way of overriding the tasks that run concurrently when serving the site with _grunt_.

This commit correct the path to **config.serve.concurrent**, so it will be able to override it.


grayside:
There's no test coverage on this either way, and it seems a very unused feature, so going with a merge.

Issue #336, opened 2017-12-01
Fix [phpcs] validation errors during the build process


scottalan:
Currently failing phpcs validation:

```
Running "phpcs:validate" (phpcs) task

FILE: ...st/working_copy/src/modules/gdt_test/tests/src/Unit/PassTest.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 1 | ERROR | [x] The PHP open tag must be followed by exactly one
   |       |     blank line
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------


FILE: /var/gdt/test/working_copy/src/modules/test.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 1 | ERROR | [x] The PHP open tag must be followed by exactly one
   |       |     blank line
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------


FILE: /var/gdt/test/working_copy/src/modules/test.yml
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 3 | ERROR | [x] Expected 1 newline at end of file; 3 found
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

PR #337, opened 2017-12-01
Issue #336 : Fix phpcs validation errors during build

scottalan:
* test/test_assets_d8/src/modules/gdt_test/tests/src/Unit/PassTest.php
  * test/test_assets_d8/src/modules/test.php
  * test/test_assets_d8/src/modules/test.yml


grayside:
These fails are on purpose.

Issue #338, opened 2017-12-01
Fix failing [eslint] tests during build process


scottalan:
Travis CI job with failing tests https://travis-ci.org/phase2/grunt-drupal-tasks/jobs/298262719

PR #339, opened 2017-12-01
Issue #338 : Fix failing eslint tests.

scottalan:
* Not really sure why these tests are here and what they do
  * Fixed error regardless


grayside:
These fails are on purpose.

PR #342, opened 2018-01-26
Only pass dbUrl to Drush if set.

### LINKED ISSUE #341: PHP notice in install task on Drush 9
While working on https://github.com/phase2/generator-gadget/pull/122, when running `grunt install`, these php notices are thrown.

```
Notice: Uninitialized string offset: 0 in vendor/drush/drush/src/Preflight/ArgsRemapper.php on line 50
```

```
Notice: Uninitialized string offset: 0 in vendor/drush/drush/src/Preflight/ArgsPreprocessor.php on line 72
```

This happens because something is passing arguments like this:

```
Running "drush:install" (drush) task
Verifying property drush.install exists in config...OK
File: [no files]
Options: cmd="/Users/jonathan.hedstrom/work/p2/gadget-test/vendor/bin/drush", cwd=false, options={"stdio":"inherit"}
Options: cmd="/Users/jonathan.hedstrom/work/p2/gadget-test/vendor/bin/drush", cwd=false, options={"stdio":"inherit"}
Args: --root=build/html, site-install, -yv, standard, , --verbose
```

note the `, ,`. This somehow translates to an empty string being passed into drush as an argument.

jhedstrom:
- Fixes #341

### DIFF
--- a/tasks/install.js
+++ b/tasks/install.js
@@ -25,10 +25,11 @@ module.exports = function(grunt) {
 
   if (dbUrl) {
     dbUrl = '--db-url=' + dbUrl;
+    args.push(dbUrl);
   }
 
   grunt.config(['drush', 'install'], {
-    args: args.concat(['site-install', '-yv', profile, dbUrl]),
+    args: args.concat(['site-install', '-yv', profile]),
     options: _.extend({
     }, cmd)
   });


tasks/install.js
```diff
@@ -25,10 +25,11 @@ module.exports = function(grunt) {
 
   if (dbUrl) {
     dbUrl = '--db-url=' + dbUrl;
+    args.push(dbUrl);
   }
 
   grunt.config(['drush', 'install'], {
-    args: args.concat(['site-install', '-yv', profile, dbUrl]),
+    args: args.concat(['site-install', '-yv', profile]),
```
grayside: @jhedstrom it seems like this means the --db-url flag would be serialized into the command before site-install, e.g., `drush --db-url=mysql://... site-install -yv standard`. Would that work?

jhedstrom: @grayside in testing drush seemed not to care (I had the same thought).

PR #343, opened 2018-02-23
Add conflict with behat >= v3.4

grayside:
Add a conflict with behat 3.4+, this is a dodge to address https://github.com/Behat/Behat/issues/1085


grayside:
The fail here relates to memory limits on composer, will power on for now and address later if it's not a fluke.

